PR #5, opened 2018-11-25
Sorted fileName and order of CreateAssetMenu for better usability/consistency

btsslawa:
Hey there,

I set up the default filenames when creating the ScriptableObjects and set their order across variables, constants, lists etc. so it looks consistently across the CreateAssetMenu-Navigation.


AdamRamberg:
Looks like some great additions! Will look further into it later this week.

AdamRamberg:
Merged! Thanks for the contribution!

Issue #11, opened 2019-02-25
[v1.0.0] Provide a Object.GetHashCode for ScriptableObject base


AdamRamberg:
See this: https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=netframework-4.7.2

Issue #10, opened 2019-02-25
[v1.0.0] AtomicTags.GetAtomicTags


AdamRamberg:
Add an GameObject extension method like this: 
`public static List<string> AtomicTags.GetAtomicTags(GameObject go)`

Issue #9, opened 2019-02-25
[v1.0.0] Cache GetComponent in extension methods


AdamRamberg:
The [extension methods](https://github.com/AdamRamberg/unity-atoms/blob/dfd70a8944acbeabe5feba1cf6cff2be6802c470/Assets/UnityAtoms/Extensions/GameObjectExtensions.cs) for AtomicTags are making calls twice to GetComponent. Cache the result from the first call.

Also go through other extension methods and make the same change.

PR #12, opened 2019-02-25
More efficient AtomicTags

### LINKED ISSUE #8: Efficient AtomicTags
I just saw the atomittacs issue/feature and it has imho many of the problems unity has with its own tags:

https://github.com/AdamRamberg/unity-atoms/blob/v1.0.0/Assets/UnityAtoms/AtomicTags/AtomicTags.cs

checking if a tag is in the list: O(n) in a hashset O(1). now, hashsets are not serializable but: [ISerializationCallbackReceiver](https://docs.unity3d.com/ScriptReference/ISerializationCallbackReceiver.html) to the rescue.

also, tagged objects could add/remove themselfs (OnEnable / OnDisable) from a `static Dictionary<StringReference, List<GameObject>>` with the addition of `public static GameObject[] FindByAtmicTag(...)`

unity has for its own tags a list of all tagged objects and iterates through all of them, IIRC.

also make AtomicTags a [ListComponent](https://garry.tv/2015/06/14/unity-tips/) or something similar... maybe: `static Dictionary<GameObject, AtomicTags>` this could help further down below on HasTag
_________________

for the [extension methods](https://github.com/AdamRamberg/unity-atoms/blob/dfd70a8944acbeabe5feba1cf6cff2be6802c470/Assets/UnityAtoms/Extensions/GameObjectExtensions.cs)

they all scream inefficient. GetTags calls GetComponentTwice -> could cache it
(same for HasTag)

HasTag could become:

    var tags = AtomicTags.GetForGameObject(go); // uses the static Dictionary<GameObject, AtomicTags>
    return tags != null ? tags.Contains(str) : false; 

**note:** also overload hashCode from StringConstant to be equal to its `Value` member. maybe do it for all in the base class ScriptableVariable base, can't think of any drawback for now.

Equivalently GetTag could become:

    var tags = AtomicTags.GetForGameObject(go);
    return tags != null ? tags.GetTags() : new List<StringConstant>(0);

soraphis:
Fixes #8 #9 #10 #11 

- Improved Efficiency of AtomicTags
- Added Equality Members for ScriptableObjectBase
- Included UnitTests for those changes
- Added C#-Assembly definitions (this had to be done because of the UnitTests)


AdamRamberg:
Hi @soraphis and thanks for the contribution! Sorry for the inactivity, been very busy lately moving to a new apartment. I will go through the pull request today or later this week.

AdamRamberg:
Just got the time to test it. Great work! Merged into v1.0.0. ⭐️ 

I'm not 100% sure what I feel about the methods: AddTag, RemoveTag and GetAtomicTags, since they all create garbage. Merged for now at least and have to think about it some more.

soraphis:
AddTag / RemoveTag will probably never be used that frequently that it matters at all. in an ideal case they are never used. i included them because this project is basically a library and it would be bad to not have them. 
Unity has all those "Find" methods, even though they are slow. but in some cases they might be usefull and if they are not used every single frame nobody really cares.

(especially when the incremental garbage collector is stable)

GetAtomicTags do not create garbage. at least not in the ususal sense. yes it allocates memory, but all it allocates is returned to be used by the programmer that wants all the tags.

if he just wants to look for a tag and then dump the list: HasTag does not allocate memory.

I think it is important to not return the List of tags, because people working with the resulting list could modify it not thinking about the changes reflecting the actual object.

my findings:

| method | GC Alloc (Bytes) |
| --- | --- |
| HasTag | 0 |
| GetAll | 32 + 8 * number of tags |
| Add | 32 + 8 * number of tags (after adding) |
| HasAnyTag | 112 (0 if only 1 element to be found)|
| Remove | 112 + 32 + 8 * number of tags |

i actually don't get why Remove allocates more than Add. in the case of Add the Sort() call does not allocate additional memory. in the case of HasAnyTag and Remove it does.

i did a few tests on HasAny e.g. checking if the amount of items in the lists were low, and then just checking all the pairs (to get rid of the sort) but the time method took went up quite a bit (in all cases, even when skipping the sort).

i also added a method to check if a list is already sorted (because C#'s sort method does not check this, it only checks if there is only 1 or less elements in it), this got rid of the GC Alloc in some cases, but the runtime went 3 times as high.

I'm thinking to add a "HasAnyTagNoAlloc" method. but i think i'm gonna wait for real use cases.

AdamRamberg:
I think you are making a good point regarding GetAtomicTags, the caller should be aware of that it allocates memory and use the method accordingly 👍 

My only use case for AtomicTags (currently) requires HasTag, so from my perspective it's all fine. My thinking though is similar to yours when it comes to AddTag / RemoveTag. Are they really needed? Do you have any use case for them? I'm thinking that the library should aim for making it easy to use in the correct way and hard to use in an unintended way (read this blog post about the "pit of success"  https://blog.codinghorror.com/falling-into-the-pit-of-success/).

soraphis:
> the library should aim for making it easy to use in the correct way and hard to use in an unintended way 

i absolutely agree with that.

> My thinking though is similar to yours when it comes to AddTag / RemoveTag

just a thought, not a real scenario:

having a tag wounded. your health component will add/remove this tag wheather its below 70% or not.    
using this and the taggedGOs to efficiently get a list of gameobjects that are wounded.

> note: i just found a bug :D and a missing feature, will create a pull request.

I've written all the stuff on top this morning without pressing "Comment".
I've rewritten the tag system completely. everything fast, no GC Alloc (one exception)

grafik

The AddCompletelyNewTag allocates memory.
I implemented ISerializationCallbackReceiver for an SortedList (seems to work, from what i've tested), i'm caching the keys of this sortedlist and wrapped those around a ReadOnlyList (an class i wrote that allows access like an list, except everything that could mutate the list)

maybe it is a bit overkill

AdamRamberg:
Awesome! Will take a look at the pull request ✌️

PR #14, opened 2019-02-26
fix warnings

rakkarage:
there are a couple ways to fix these warnings
this is one
anything is preferable to 10+ warnings?
screenshot from 2019-02-25 17-30-06

### DIFF
--- a/Assets/UnityAtoms/CreateOnAwake/CreateListOnAwake/CreateListOnAwake.cs
+++ b/Assets/UnityAtoms/CreateOnAwake/CreateListOnAwake/CreateListOnAwake.cs
@@ -21,16 +21,16 @@ namespace UnityAtoms
         private bool CreateClearedEvent = false;
 
         [SerializeField]
-        private TEL AddedListener;
+        private TEL AddedListener = null;
         [SerializeField]
-        private TEL RemovedListener;
+        private TEL RemovedListener = null;
         [SerializeField]
-        private VoidListener ClearedListener;
+        private VoidListener ClearedListener = null;
 
         [SerializeField]
-        private GA1 OnListCreate;
+        private GA1 OnListCreate = null;
         [SerializeField]
-        private GA2 OnListCreateWithGO;
+        private GA2 OnListCreateWithGO = null;
 
         void Awake()
         {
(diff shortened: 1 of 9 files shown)


AdamRamberg:
Hi @rakkarage ! Sorry for the inactivity, been very busy lately. 

Why did you close the issue? I don't get the same warnings in my console.

rakkarage:
ops sorry, np thanks
i guess warnings are just in (.NET 4.x Equivalent, Mono, .NET Standard 2.0) project settings or new version or something like that

soraphis:
yeah, those warnings will appear if using the .NET 4.x in Unity, they come because unity is now using the roslyn compiler

there already is an issue for that: https://issuetracker.unity3d.com/issues/serializedfield-fields-produce-field-is-never-assigned-to-dot-dot-dot-warning

but it wont be resolved, because its "by design", nevertheless there is an issue opened on the Roslyn repository: https://github.com/dotnet/roslyn/issues/30172 regarding exactly that issue.

I'm unsure whether i want to initialize all those fields with redundant default values or not. While I myself always hide warnings in unity (because most of the time they are from a 3rd party library because of things that have been marked obsolete or somthing like that) on the other hand this project is designed to be one of those 3rd party librarys and as such we should not contribute to generate warnings in the editor.

and since such an issue on the roslyn compiler can take up to a year and then waiting for the next unity version to be used its probably for the better to add a `= default;` to all those fields.

AdamRamberg:
@soraphis Good points! Don't think the version of C# used in Unity supports setting defaults though, so nulls are fine. 

@rakkarage Could you do these changes in the v1.0.0 branch instead? v1.0.0 is the currently active branch and I'm planning to merge that into master pretty soon. On another note, think we got some different settings in our editors, because the indenting in this commit is different from mine. Just committed an .editorconfig to fix that (v1.0.0 branch). I will close this pull request and would happily accept a pull request in branch v1.0.0 with the fixes mentioned above ⭐️

AdamRamberg:
FYI. Fixed the warnings in v1.0.0

rakkarage:
thanks...
another way to 'fix' is to just make stuff you want editor to edit, **public**
less code / attributes and makes sense too imo

AdamRamberg:
I think it is better to explicitly set variables to private and use the SerializedField attribute though. It makes it more clear that this is a variable that should only be public in the sense that you are going to inject a value from the Inspector.

PR #15, opened 2019-03-07
More AtomicTags changes

soraphis:
- Removed the Plugins.meta file i accidently added
- Complete AtomicTag rewrite, as GC alloc free as possible:

grafik

- Even better tested!


AdamRamberg:
Looks great! +1 for the tests, didn't know that you could profile inside a unit test, great stuff ⭐️ 

Will merge now, but in the future could you use omnisharp for code formatting so that the formatting is consistent between committers? Or if you got some other idea / solution for that? I'm a web dev by day so really found of ESLin, from what I found so far Omnisharp's code formatting is the closest thing to that.

Issue #21, opened 2019-03-28
Generator to create new Atoms (classes)


AdamRamberg:
It would be really nice to have a generator that creates new atoms classes for you, to reduce the amount of boilerplate code to be written. It should be possible to choose what to generate (Variable, EventListener, Constant, etc), what to name it and what types to use. 

Not sure what techniques to use. Have worked with Yeoman / EJS before, but not sure if it's a good fit for Unity / C# (is it possible to run from the inspector if it's a node module for example?). Maybe do some custom templating?

soraphis:
in my fork, i already added a code generator

https://github.com/soraphis/unity-atoms/blob/stable/Editor/Dev/CodeGenerator.cs

this is ignored for package manager, and it uses my refactored folder structure, but you might find it helpfull

AdamRamberg:
@soraphis That is great! Would you mind creating a pull request for this? ⭐️ 

Think it shouldn't be ignored by the package manager though. It should be possible for users of Unity Atoms to easily generate their own atoms.

AdamRamberg:
Initial generator has been mereged from @soraphis branch. Closing this one and will create a new issue for future improvements.

PR #28, opened 2019-04-05
Code Generation

soraphis:
addressing #21 

For a better overview I moved all the Type-Folders into an parent directory "Types".
And implemented the Code generator from my fork.
I tested it with Vector4 which worked just fine. 

To access the code generator, in the menu bar should be an option "Tools > UnityAtoms > Code Gen Utility". this opens a window. where you can enter a path and and a typename.

This window is written with the Experimental "UIElements", so at some point we need to remove the "experimental" imports. We'll see when Unity 2019.1 or 2019.2 are released


AdamRamberg:
@soraphis Can you turn on "Allow edits from maintainers"? Would like to make some minor changes before I merge

soraphis:
@AdamRamberg it already is

Issue #25, opened 2019-03-30
Make Variables discard playmode changes


soraphis:
this week i was working on small project and we used the variables of unity-atoms pretty extensively.

And one of the more annoying this is, that scriptable objects keep playmode changes in the editor.
So you have to write scripts that setup your variables to start values.

I propose to make the Variable classes Implement ISerializationCallbackReceiver like shown here: https://unity3d.com/de/how-to/architect-with-scriptable-objects (the 3rd image)

soraphis:
could be implemented like this: https://github.com/soraphis/unity-atoms/commit/f95307c8068dde0a6994eaf0cec4d67761f01ad4

AdamRamberg:
This is a great feature and something that I've been thinking about / toying around with myself. However, I think there are different use cases, sometimes you want the value to persist between scenes and plays and sometimes you want it to be reseted to an initial value. Therefore, I propose that the ScriptableObjectVariable<T> should also contain a ScriptableObjectVariableBase<T> called initalValue (this will need to be part of the generic signature in order to be serialized properly). If the constant (ScriptableObjectVariableBase) is set we set in OnAfterDeserialize, otherwise we don't. Would gladly accept a pull request (but I might also try to implement this right away - so check if it has already been implemented before doing a pull request).

Side note: when you want to reset something at the start of a scene you can use OnStartHook combined with SetVariable.

soraphis:
> sometimes you want the value to persist between scenes and plays

this is a dangerous thing! while you might want it for in-editor features this will never happen in builds. because of that I highly recommend implementing a reset system.
(also: in-editor features will probably not use unity-atoms)

> Therefore, I propose that the ScriptableObjectVariable should also contain a ScriptableObjectVariableBase called initalValue (this will need to be part of the generic signature in order to be serialized properly)

I'd argue against that, because: you'd need for nearly all variables an additional variable for resets. this reset variable would have its own description fields, old values, events, and so on

while all you actually want is: a boolean flag whether or not you wanna reset, with a default on true. (because really, I can't imagine a case where you want to store something inbetween editor sessions)

> Side note: when you want to reset something at the start of a scene you can use OnStartHook combined with SetVariable.

which will introduce WAY MORE assets than i actually need (it doubles the amount).

in the game we made the other week, we had around ~10!? unity-atom variables. that would be 20 + the OnStartHook.

AdamRamberg:
> while you might want it for in-editor features this will never happen in builds. 
[...]

PR #29, opened 2019-04-05
scriptablevariables now discard playmode changes.

soraphis:
addressing #25 

Implemented ISerializationCallbackReceiver like its shown here: https://unity3d.com/de/how-to/architect-with-scriptable-objects

### DIFF
--- a/Source/Base/Variable/ScriptableObjectVariable.cs
+++ b/Source/Base/Variable/ScriptableObjectVariable.cs
@@ -8,7 +8,7 @@ namespace UnityAtoms
         where E1 : GameEvent<T>
         where E2 : GameEvent<T, T>
     {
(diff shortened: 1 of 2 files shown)


AdamRamberg:
Thanks for the PR!

Merged in 06eab90bb184d9cebb6d39336c7dd4bb391407e1 and fixes in ed574215ad4e09a7cc9e6b494da9691d26786e37

Don't see a reason for it to be called runTimeValue, so renamed it back to value. There were a couple of issues with SerializeField as well that I fixed. The major adjustment I had to do was to move the OnAfterDeserialize to ScriptableObjectVariable instead of ScriptableObjectVariableBase, since we don't want an inital value for constants.

PR #30, opened 2019-04-07
Code cleanup

jeffcampbellmakesgames:
### Summary
This PR takes a first stab at some of the low hanging fruit for code quality by:

* Cleaned up line-endings across the solution (all line endings converted to Mac/Linux style (LF) to match editor config) and added new-lines at the end of files that did not have them.  
* Added .gitattributes file with check to ensure line endings for all csharp files are converted to LF when checked in. This should help prevent mixed OS line endings.
* Removed unused namespaces across all files in sample, test, and atoms runtime/editor code.
* For parameters on methods that hid existing members, these were renamed so that they did not do that.

### Testing
There are no functional changes associated with this PR, but in general running unit-tests and playing the sample content should be enough.


soraphis:
While in general I think those changes are nice, I'm unsure about the coding conventions, especially because there are differents between Unity and .NET

here is a comparison https://gist.github.com/soraphis/7651bc5a8bb8e26fd5aa81b239aee835

AdamRamberg:
Thanks for the PR! 

The most important thing is to find a style guide that makes the code base consistent through out the project. Looking at this one: https://gist.github.com/soraphis/7651bc5a8bb8e26fd5aa81b239aee835 it seems like they are the same until we get to ”public field names”. When it comes to public fields I would say that we should opt-in to use PascalCase, since only time you really do this is when you are too lazy to make a property (it should be available via the class’s API). I really don’t like using ”m_” (hungarian notation), it looks cluttered and doesn’t really add anything (if you don’t use it for all your class scoped variables / fields). I also agree that constants should use snake case (upper case). I also don’t see why you would differ read only and static fields compared to regular private fields.

Think we can argue about this for a long time, so I'm just going to decide that we will use the below guidelines: 

Object Name | Notation | Example
-- | -- | --
Namespaces | PascalCase | UnityAtoms
Class names | PascalCase | AtomicTags
Method name | PascalCase | GetTags
Method arguments | camelCase | oldValue
Property names | PascalCase | Value
Public Field names | PascalCase | Value
Private Field names | _camelCase | _value
Private / public constants | SNAKE_CASE | NETWORK_ACCESS_TOKEN_SIZE
Inline variables | camelCase | value
-- | -- | --

I will make adjustments to this PR accordingly and merge it.

AdamRamberg:
@jeffcampbellmakesgames could you check "Allow edits from maintainers"? Would like to commit some small fixes before I merge.

AdamRamberg:
Merged in b54bf46bff24c3537ef9f20fb783a1ec77a08c46

Jeffrey-Campbell-Zapdot:
Overall I think these changes look reasonable. Unitys standard for disambiguating property and field names is not great to go by. The reason I would advocate for using "PascalCase" for properties and "pascalCase" for fields is that they are not just semantically different; calling a property is much more akin to calling a method than just a getter/setter. Events might be invoked, memory allocated, etc... 

For example, one thing I've seen trip people up is over-use of GameObject.name, which is actually a property that allocates garbage when used. See [here ](https://unity3d.com/learn/tutorials/topics/performance-optimization/optimizing-garbage-collection-unity-games) for more details, but the important part of the text is merged below.

> Another unexpected cause of heap allocations can be found in the functions GameObject.name or GameObject.tag. Both of these are accessors that return new strings, which means that calling these functions will generate garbage. Caching the value may be useful, but in this case there is a related Unity function that we can use instead. To check a GameObject’s tag against a value without generating garbage, we can use GameObject.CompareTag().

> In the following example code, garbage is created by the call to GameObject.tag:

```
private string playerTag = "Player";

void OnTriggerEnter(Collider other)
{
    bool isPlayer = other.gameObject.tag == playerTag;
}
```

> If we use GameObject.CompareTag(), this function no longer generates any garbage:

```
private string playerTag = "Player";

void OnTriggerEnter(Collider other)
{
    bool isPlayer = other.gameObject.CompareTag(playerTag);
}
```

That `tag` or `name` isn't just a string value isn't obvious when looking at in unless you've anecdotally learned that. That being said I could probably live with that, but I do believe its a bad practice.

> @jeffcampbellmakesgames could you check "Allow edits from maintainers"? Would like to commit some small fixes before I merge.

Sure thing.

Jeffrey-Campbell-Zapdot:
Not quick enough! I wish I typed faster :/

AdamRamberg:
No worries, I just wanted it to be merged so I created a new merge branch instead :)

AdamRamberg:
One thing though, cause I'm a little bit confussed. Is this containing the same changes as PR #31?

Jeffrey-Campbell-Zapdot:
Yes, that can be closed; I had meant to PR this in a feature branch to `canary` but had accidentally done it from my forked `canary`. That one can be closed.

AdamRamberg:
Alright, got you 👍

PR #31, opened 2019-04-08
Code quality, style, and formatting (Request for Comments)

jeffcampbellmakesgames:
### Summary
I took some time this weekend reading through the codebase at large and decided to try and cleanup inconsistencies in code style, restructuring code for readability, and applying some best practices for Unity at large. I've also updated the readme to include a style guide section [here](https://github.com/jeffcampbellmakesgames/unity-atoms/tree/fix/code_quality#style-guide) that covers this style for contributors.

### Changes
Cleaned up code readability, naming style to be consistent

* Add abstract modifier to all generic classes meant not to be instantiated directly as the intent is to specify a pattern for a closed type that can be serialized.
* Added sealed keyword to all closed types. This is an incidental performance improvement for mobile and desktop platforms utilizing il2cpp due to the way unsealed vs sealed virtual methods are handled in generated C++ code.
* Also sealed all inspectors to mark them as closed types.
* Dropped all where constraints to the next line as most of these class declarations break 120-130 characters. This should help improve readability and make this style consistent as it has been in use for the most-complex generic types already, but not for others.
* Dropped all generic type parameters (generally) numbering three or more to the next line to improve readability.
* Where found, added empty lines between property/field declarations in classes to improve readability.
* Extracted several 2x event classes into their own files to match convention established by other event classes.
* Removed unecessary meta files for C# solution.
* Added compiler options file (csc.rsp) file for the Unity project and added global warning ignore for 0649 (Field 'field' is never assigned to, and will always have its default value 'value'). This is necessary since 2018.3 switched the compiler to Roslyn and by design they have decided to not suppress this warning for monobehavior fields that are private, but serialized. This is very widely contested as most developers will not make every field needed to be assigned in the Editor as public (breaks encapsulation, opens surface area for bugs) and this can in some cases generate hundreds if not thousands of warnings. Adding the compiler options file suppresses this warning which also may hide legitimate warnings, but is far the lesser of two evils.
* Moved example scripts not in a namespace to UnityAtoms.Examples.
* Reordered public/private fields and properties to be consistent. Order is as follows; public and private properties followed by public/private fields.
* Renamed private fields to use '_' prefix and marked public fields as private where only used internally. Marked these fields with FormerlySerializedAs attribute to preserve older serialized values already in use.
* Removed redundant initialization of null to reference types as this is their default value.
* Marked implicitly private methods and members without an access modifier as explicitly private.
* Updated unit tests to use new private name field when getting private member field info.

### Testing
There are few functional changes associated with this PR (internal changes for loop control in AtomicTags), but in general running unit-tests and playing the sample content would be good. I would also make sure to read the style guide [here ](https://github.com/jeffcampbellmakesgames/unity-atoms/tree/fix/code_quality#style-guide).


soraphis:
this seems to be a duplicate of #30 

✔️Add abstract modifier to all generic classes...
✔️ Added sealed keyword to all closed types....
❌ Dropped all where constraints to the next line as most of these class declarations break 120-130 characters.
❌ Dropped all generic type parameters

    I personally think a strictly forced line width does more harm than use.
    e.g:

    public abstract class ScriptableObjectList<T, E> : ScriptableObject where E : GameEvent<T>
    vs
    public abstract class ScriptableObjectList<T, E> : ScriptableObject
        where E : GameEvent<T>

    or:

    public class BoolReference : ScriptableObjectReference<bool, BoolVariable, BoolEvent, BoolBoolEvent> { }
    vs 
    public sealed class BoolReference : ScriptableObjectReference<
        bool,
        BoolVariable,
        BoolEvent,
        BoolBoolEvent> { }

🆗 added empty lines between property/field declarations in classes to improve readability
🆗 Extracted several 2x event classes into their own files to match convention established by other event classes

    I personally thought the complete oposite: since most of those classes are empty, they could share the same file. Their only use is to define the generic types.

✔️ Removed unecessary meta files for C# solution
🆗 Added compiler options file (csc.rsp) file for the Unity project and added global warning ignore for 0649

    but this will not solve the problem for people using UnityAtoms. (correct me if I'm wrong)

✔️ Moved example scripts in a namespace to UnityAtoms.Examples.
🆗 Reordered public/private fields and properties to be consistent. Order is as follows; public and private properties followed by public/private fields.
❌ Renamed private fields to use '_' prefix and 

see my comment in #30 https://gist.github.com/soraphis/7651bc5a8bb8e26fd5aa81b239aee835

especially the "private static" -> PascalCase does not make sense to me. Imho it should follow the other "private fields"

✔️ marked public fields as private where only used internally. 
✔️ Marked these fields with FormerlySerializedAs attribute to preserve older serialized values already in use.
✔️ Removed redundant initialization of null to reference types as this is their default value.
✔️ Marked implicitly private methods and members without an access modifier as explicitly private.
✔️ Updated unit tests to use new private name field when getting private member field info.

_______________

> internal changes for loop control in AtomicTags

why? this should not make any difference in this case, because the ReadonlyList is just returning the Enumerator of the List it references, and iterating over List<T> should not allocate memory - at least not since Unity 5.6

PR #44, opened 2019-06-17
Canary/generator fix

soraphis:
This PR fixes: 

Generator.cs must not be outside of the Editor Assembly Definition because it uses UnityEditor and thus would prevent games from building

Cleaned up `using` statements in generator related classes.

Removed unused methods in Generator code

Fixed a potential bug (and at least resolving warnings) because IndexOf is culture specific
___________

If I get this right, Templating.cs will fall apart when there are nested `<%IF ... > ` blocks.
It would be smarter to use an recursive approach


soraphis:
Yeah I was right about the thing with the nested-IF blocks.

I added a fix for that. the results can be checked here: https://repl.it/@soraphis/CautiousFocusedMicrocode

AdamRamberg:
Thanks @soraphis ! Nifty implementation of the templating algorithm! ⭐️ 

Got one question, even though I closed the PR. Have you checked that you allow the maintainer to push commits to the PR? I tried to push some small style tweaks in 2 commits to your PR, but got permission denied. It's really annoying. I might be doing something weird, but don't thinks so. I run the following commands: 
```
git checkout -b soraphis-canary/generator-fix canary
git pull https://github.com/soraphis/unity-atoms.git canary/generator-fix
```
And then make my code changes. Then: 
```
git add .
git commit -m "My changes"
git push https://github.com/soraphis/unity-atoms.git soraphis-canary/generator-fix
```
Then I'm getting: 
` ! [remote rejected] soraphis-canary/generator-fix -> soraphis-canary/generator-fix (permission denied)`

This is driving me nuts 😨

soraphis:
> Have you checked that you allow the maintainer to push commits to the PR?

Github seems strange in this regard. when creating the PR there is the "allow maintainer to push" checkbox which is ticked. but after creating the PR it seems the checkbox on the right side gets disabled.

I now checked it, and I'll make sure in future to double check this after creating the PR.


Another thing in that regard you might want to consider, instead of pushing yourself to canary directly creating feature branches and creating PRs yourself, giving others some time (a day or two) to give a code review. (I mean its your repo and you can do as you like, its just a suggestion)

AdamRamberg:
Hmm that is weird 🤔Would be great if you really make sure next time to check it! 

I understand that it would be desirable that I create PRs. However, I've got a bunch of changes that I want to get through here pretty soon in order to release v2. I will reconsider this after I have released v2 / or before that but after all the changes I have in mind are done.

Issue #45, opened 2019-06-24
Unity Atoms unable to find UnityEngine.UI


CapitanLiteral:
I'm using 2019.3.0a4 version

When I add atoms on package manager this i what I see on my console, any Idea of what could be happening?

[CompilerError] The type or namespace name 'UI' does not exist in the namespace 'UnityEngine' (are you missing an assembly reference?)
Compiler Error at Library\PackageCache\com.mambojambostudios.unity-atoms@163a216fc14abf9b9f6014ec77ea1fe31c547c4f\Source\MonoHooks\OnButtonClickHook.cs:2 column 19
1:   using UnityEngine;
-->2:   using UnityEngine.UI;
4:   namespace UnityAtoms

[CompilerError] The type or namespace name 'EventSystems' does not exist in the namespace 'UnityEngine' (are you missing an assembly reference?)
Compiler Error at Library\PackageCache\com.mambojambostudios.unity-atoms@163a216fc14abf9b9f6014ec77ea1fe31c547c4f\Source\MonoHooks\OnPointerDownHook.cs:2 column 19
1:   using UnityEngine;
-->2:   using UnityEngine.EventSystems;
4:   namespace UnityAtoms

[CompilerError] The type or namespace name 'IPointerDownHandler' could not be found (are you missing a using directive or an assembly reference?)
Compiler Error at Library\PackageCache\com.mambojambostudios.unity-atoms@163a216fc14abf9b9f6014ec77ea1fe31c547c4f\Source\MonoHooks\OnPointerDownHook.cs:7 column 55
5:   {
6:       [AddComponentMenu("Unity Atoms/Hooks/On Pointer Down")]
-->7:       public sealed class OnPointerDownHook : VoidHook, IPointerDownHandler
8:       {
9:           public void OnPointerDown(PointerEventData eventData)

[CompilerError] The type or namespace name 'PointerEventData' could not be found (are you missing a using directive or an assembly reference?)
Compiler Error at Library\PackageCache\com.mambojambostudios.unity-atoms@163a216fc14abf9b9f6014ec77ea1fe31c547c4f\Source\MonoHooks\OnPointerDownHook.cs:9 column 35
7:   public sealed class OnPointerDownHook : VoidHook, IPointerDownHandler
8:   {
-->9:       public void OnPointerDown(PointerEventData eventData)
10:       {
11:           OnHook(new Void());

[...]

PR #46, opened 2019-06-27
Unity Atoms 2019.2+ support

soraphis:
fixing #45 where in newer unity versions all ugui stuff is in an own package, which needed to be added in the assembly defintion.

**note**: I'd rather add this changes to its own branch (e.g. canary-2019.2 or whatever) because it could give problems with current stable unity versions (2018.3 or the 2018.4 LTS when it arrives) 

but when creating a pull request github does not allow to create a new target branch

### DIFF
--- a/Source/UnityAtoms.asmdef
+++ b/Source/UnityAtoms.asmdef
@@ -1,3 +1,14 @@
 {
-	"name": "UnityAtoms"
+    "name": "UnityAtoms",
+    "references": [
+        "GUID:2bafac87e7f4b9b418d9448d219b01ab"
+    ],
+    "includePlatforms": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": []
 }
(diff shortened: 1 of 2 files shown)


AdamRamberg:
Thanks for the PR @soraphis! 

To be honest, I'm not completely sure on how the .asmdef files work. What is the implication of adding it? Am I understanding it correctly if I say that it gives direction on what packages / references Unity Atoms needs in order to run?

soraphis:
> To be honest, I'm not completely sure on how the .asmdef files work

It makes a separate assembly for all code files in its folder or subfolder (that don't have their own assembly definition).

In this assembly definitions you can reference other assemblies. you need to do that to reference classes from those other assemblies. 

since unity moved the ugui stuff to its own assembly, and we're referencing ugui stuff, so we need to depend on that assembly. 

_____________

splitting into assemblies has for example the benefit of reducing compile time (because only the assembly with the changes needs to be recompiled). So for libraries you absolutely want an assembly definition to not pull down compile times for users.

AdamRamberg:
@soraphis Thanks for taking the time to explain the problem! ⭐️ 

I will accept the PR, hopefully it won't break stuff in older version of Unity Atoms (but that what's the canary is branch for, ey?). Will work on Unity Atoms today and hopefully I will get to the new project structure and thereby I will create separate packages for Unity UI, Tags, etc. When done I will add the assembly definition to Unity Atoms UI instead of to the full project.

Issue #47, opened 2019-07-14
Build error 2019.1.5f1 - MonoScript, AssetPostprocessor could not be found; Canary branch


IceTrooper:
Hi, I got a problem with the build in Unity 2019.1.5f1 (platform doesn't matter; I tried both on Android build and Windows build).
More information in the screenshot. Is it possible to get a fix?

BTW: This is my first project with UnityAtoms and until now, it works really nice. It takes some time to get used to, some things could be better, but I really appreciate your repo. A great extension to Unity workflow. Thank you
Do you consider to enhance this repo with new features? Or show it on Unity forum (I haven't found anything related to that repo)?

![unityatoms_error](https://user-images.githubusercontent.com/51023959/61189336-4254d980-a68c-11e9-950f-4a5fb0dd5f4d.png)

soraphis:
> More information in the screenshot. Is it possible to get a fix?

I've had a look into that. it should be an easy fix. i'll make a PR

> Do you consider to enhance this repo with new features? Or show it on Unity forum (I haven't found anything related to that repo)?

Feel free to leave feature requests as issues! The project is currently in a kinda reconstruction phase, because of uncertainty regarding the package system. but yes: it is planned to enhance and update the project

IceTrooper:
Great! I will be waiting for a fix. I just wanted to test my project on phone, heh.
I'm advertising that repo in Discord GameDev server every time when Unite 2017 talk or ScriptableObjects topic comes. I would like to see more people use it ^^

AdamRamberg:
> Do you consider to enhance this repo with new features? Or show it on Unity forum (I haven't found anything related to that repo)?
Yes, like @soraphis was saying, we plan to continue with improvements! Thought that I made a post about it on the Unity forums the first time I released it, but can't find that post now. Will take a look at it and might create a new one if needed. 

@soraphis Thanks for the quick PR! ⭐️

AdamRamberg:
Fixed in #48

PR #48, opened 2019-07-14
Fixing some classes requiring UnityEditor assembly while being outside of Editor asmdef

soraphis:
Moved a few classes from IconAssignment into Editor assemblies because they require UnityEditor

fixing #47


AdamRamberg:
Looks good! Thanks @soraphis  ⭐️

PR #57, opened 2019-10-08
Add Collision, Collision2D and object Variables

lucasrferreira:
Collision and Collision2D are analogue to Collider Variables and Events.

'object' is useful when an 'wildcard' variable is needed.


AdamRamberg:
Thanks for opening the PR! 

I actually agree with @soraphis. I don't think that this should be part of the core package. If you want these types in your project you should be able to easily generate these types using the generator: https://adamramberg.github.io/unity-atoms/introduction/generator

If you really want, and if you think there is a need for it, you could make your own sub package for Unity Atoms and redestribute it on Github :) 

That said I really appreciate you making the effort to submit a PR. Before you submit your next PR please also read the [guidelines](https://github.com/AdamRamberg/unity-atoms/blob/canary/CONTRIBUTING.md). Closing the PR.

Issue #69, opened 2019-10-29
[BUG] Editing Value of Unity Atom in Editor immediately reverts value to Initial


Sparkky:
**Describe the bug**
You cannot edit the value of a variable in the editor  as soon as you change focus the value immediately reverts to Initial Value, furthermore if you do this while the game is playing, no update event will fire.

**To Reproduce**
1. Create blank project with just Unity Atoms 2.0
2. Create a Unity Atom (Int for this example)
3. try to change "Value" of the Atom
4. as soon as you change focus / hit enter the value will revert to Initial value

**Expected behavior**
You can change value in editor and the value will change, Ideally if there is an associated event and the game is running it should fire that event.

**Environment**
 - OS: windows 10
 - Unity Version 2019.2.9
 - Unity Atoms v 2.0.0

**Additional context**
If the value was changed by another script, everything works as expected, you can watch the value update in the editor drawer.

soraphis:
if not in playmode: change the initial. 

if in playmode: when you're changing things directly in editor this is done via unitys serialization system it does not call the setter method and thus it does not send events. on the event there is a button to manually send an event from the editor.

this is expected behavior.

though: this could be made clearer for the user. will look into this

AdamRamberg:
Thanks for submitting this issue! 

Elaborating on what @soraphis is saying, we should definitely look this over. In my mind the optimal UX would be if: 
- **Value** - always editable both in edit mode and play mode. If possible it should trigger a change event when changed in play mode. 
- **Initial Value**  - always editable both in edit mode and play mode.
- **Old Value** - has not been part of the discussion above, but want to add this to the discussion since it is still a part of the UX of the Variable's editor. I think this one should be visible , but never editable. Maybe we should add something like [this](https://answers.unity.com/questions/489942/how-to-make-a-readonly-property-in-inspector.html). Should we create a separate issue for this? 

To solve the immediate problem described by @Sparkky I think it would be ok to move the assignment of `_value` to `_initialValue` from `OnAfterDeserialize` to the beginning of `OnEnable`. This at least works great for me when running in the editor, but not 100% sure that this behaves correctly in builds.

soraphis:
I'm working on a Custom editor:

- Old Value is never editable (does not make sense)
- Value is only ediable in playmode and triggers a change event.
- Initial Value is only ediable in edit mode (and also changes value along)

[...]

PR #70, opened 2019-10-31
Better user guidance for working with AtomVariables

soraphis:
*Problem description*

It can be confusing if and when initialValue or Value or OldValue can and should be edited. also events don't fire in runtime when the value changes.

addressing #69 

*Solution* 

Changes contain a Custom Editor (including generation template) which changes the inspector in the following way:

image

image

- Old Value is never editable 
- Value is only ediable in playmode and triggers a change event.
- Initial Value is only ediable in edit mode (and also changes value along)
   - initial value in playmode _can_ be unlocked. changing the value will use unitys serialization system and will surpass the event handlers

### DIFF
--- a/Packages/Core/Editor/Editors/AtomVariableEditor.cs
+++ b/Packages/Core/Editor/Editors/AtomVariableEditor.cs
@@ -1,0 +1,39 @@
+using UnityEditor;
+
+namespace UnityAtoms.Editor{
+    public abstract class AtomVariableEditor<T> : UnityEditor.Editor
+    {
+        protected virtual T FromProperty(SerializedProperty property) => property.GetValue<T>();
+        protected abstract T Field(string label, T value);
+
+        public override void OnInspectorGUI()
+        {
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("_developerDescription"));
+            EditorGUILayout.Space();
+
+            EditorGUI.BeginDisabledGroup(EditorApplication.isPlayingOrWillChangePlaymode);
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("_initialValue"));
+            EditorGUI.EndDisabledGroup();
+
+            EditorGUI.BeginChangeCheck();
+            EditorGUI.BeginDisabledGroup(!EditorApplication.isPlaying);
+            T value = FromProperty(serializedObject.FindProperty("_value"));
+            value = Field("Value", value);
+            EditorGUI.EndDisabledGroup();
+            if (EditorGUI.EndChangeCheck() && target is AtomBaseVariable atomTarget)
+            {
+                atomTarget.BaseValue = value;
+            }
+
+            EditorGUI.BeginDisabledGroup(true);
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("_oldValue"));
+            EditorGUI.EndDisabledGroup();
+
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("Changed"));
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("ChangedWithHistory"));
+
+            serializedObject.ApplyModifiedProperties();
+
+        }
+    }
+}
--- a/Packages/Core/Editor/Editors/AtomVariableEditor.cs
+++ b/Packages/Core/Editor/Editors/AtomVariableEditor.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Reflection;
 using UnityEditor;
 using UnityEngine;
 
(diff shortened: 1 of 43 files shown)


AdamRamberg:
This is great stuff! The changes to the Editor is really needed. There is some stuff that we need to address in this PR though. 

First of all is that when you change the `Value` in the inspector in play mode the event gets raised, but the value remains the same. This is due to the `OnAfterDeserialize` in `AtomVariable<T>`. This is something that we need to address. My suggestion is to use `OnEnable` instead since that is only triggered once when entering play mode (when the ScriptableObject gets loaded), which is the behaviour we want. `OnAfterDeserialize` on the other hand is called when starting play mode / the game, but also when Unity is serializing the object, which it does every time we do changes in the editor. 

My other main concern is the `ValueLayoutField` and `GetValue<T>` functions in `SerializedPropertyExtensions`. Since they are not truly generic these are not going to work very well with custom classes / structs, if we for example would regenerate the `TouchUserInput` Atoms it would not work. I think it is important that this functionality works with custom classes / structs being generated using the Generator. I know that getting a value from the `SerializedProperty` when not knowing the type is pretty rough though in Unity. However, I found [this](https://forum.unity.com/threads/get-a-general-object-value-from-serializedproperty.327098/#post-2309545) solution that might help us. 

To fix the 2 problems above I locally tried to the following: 
- Moved / copied `GetTargetObjectOfProperty` and `GetValue_Imp` from the link above into the project. 
- Used `PropertyField` and `GetTargetObjectOfProperty` in `AtomVariableEditor<T>` instead of `FromProperty` and `Field`. 
- Moved the code in `OnAfterDeserialize` to `OnEnable` in `AtomVariable<T>`. 
- Remove `SerializedPropertyExtensions`

The main issue with the above is that events are not triggered when changing the `Value`. We could fix that if we added a flag to `SetValue` that by passed the `AreEqual` check, eg. `byPassEqualCheck`. Then call that function in the `AtomVariableEditor<T>`. Note that this would require a change in the generic "signature" of `AtomVariableEditor<T>` to also take Event and Event x 2. 

In summary, I'm open to other solutions than mine. However, we need to solve the editor in a generic was so that it works for all types (also user created classes / structs) with potential custom property drawers.

soraphis:
> To fix the 2 problems above I locally tried to the following:
> 
> * Moved / copied `GetTargetObjectOfProperty` and `GetValue_Imp` from the link above into the project.
> * Used `PropertyField` and `GetTargetObjectOfProperty` in `AtomVariableEditor<T>` instead of `FromProperty` and `Field`.
> * Moved the code in `OnAfterDeserialize` to `OnEnable` in `AtomVariable<T>`.
> * Remove `SerializedPropertyExtensions`

*1 and *2 done.
*3 I'm still unsure about this, because I can't believe that they are equivalent. but I have not the time yet to do further testings (especially in builds). Moving it is not required anymore anyways. since the editor script just does not call `ApplyModifiedProperties` when _value was changed

 *4 since the code for (*1 / *2) this might come in handy in other editor extension later on, so I'd leave this in an Extension-Class for now

AdamRamberg:
Nice! This will still not play nicely with structs as a value though :( Will investigate further if it is possible to do this with unknown but serializable structs / classes (think I have an idea where you can traverse the `SerializedProperty`). The `SerializedProperty`class is a little bit tricky to work with though. 

**Other then that here are some additional stuff that needs to be in place:** 
- Add summary C# XML comments to the generated editors
- `ManagedReference` in `SerializedPropertyExtensions` seems to have been added lately to Unity. Add a Unity version flag. 
- Generate Editor for all types. 
- Remove unreachable code (line 47) in `SerializedPropertyExtensions.cs
- Include children `true` using `PropertyField` - in order to display structs for example.

soraphis:
> Other then that here are some additional stuff that needs to be in place:
> * Add summary C# XML comments to the generated editors
> * ManagedReference in SerializedPropertyExtensions seems to have been added lately to Unity. Add  a Unity version flag.
> * Generate Editor for all types.
> * Remove unreachable code (line 47) in `SerializedPropertyExtensions.cs
> * Include children true using PropertyField - in order to display structs for example.

- added XML comments
- removed managedReference. it used the default case anyways
- did small tweak in Generator, such that it does not use hardcoded paths, and regenerated the classes
- used includeChildren 

i think it should work for basically any serializable type (not tested yet)
I think its also possible to get rid of all the implementation classes and using the abstract base-VariableEditor class using it as 'fallback'.

needs more investigation

AdamRamberg:
Haven't tested it yet, but if the latest commit didn't add anything specific to deal with structs (eg. TouchUserInput) there is still some work to do. I experimented with this locally and think I have a solution to this. Basically you will need to traverse the SerializedProperty and set its SerializableProperties. This however, makes it impossible to just set the `BaseValue` for the AtomVariable as we do now. So have to do some other workarounds to get events and such to trigger.

If you haven't already could you please turn on to accept pushes / commits by maintainers to the PR?

soraphis:
> Haven't tested it yet, but if the latest commit didn't add anything specific to deal with structs

i don't know what you're talking about, but the ProperyField already handles structs:

image

this is for a:

```c-sharp
[System.Serializable]
public struct SuperStruct
{
    public int ValueInt;
    public float ValueFloat;
    public Vector4 ValueVec4;
    public MonoBehaviour ValueComponent;
}
```

where i generated the matching AtomTypes

soraphis:
now supports structs

soraphis:
now also addressing #72 

image

works with the same SuperStruct as above

AdamRamberg:
Made some upstream changes to this PR in order to fix the last things: 
- Call `serializedObject.Update();` before `OnInspectorGUI` in the editor script in order to update old value when changing value. 
- Show children for PropertyFields for initial and old value. 
- Some minor refactoring - renaming variables
- Clean up in `RegenereateAllAtoms`
- Use `OnEnable` instead of `OnAfterDeserialize` when setting inital value for variable. `OnAfterDeserialize` was causing major issues with the custom editor, eg. it did run when focusing a Variable and showing the inspector, basically making it worthless to change or even inspect the Variable at runtime. `OnEnable` is only called ones when the ScriptableObject is loaded in runtime, which is what we really want. This has been tested running in the editor and in builds. 
- Add missing editors scripts `TouchUserInputVariableEditor`and `SceneFieldVariableEditor`

Issue #79, opened 2019-11-10
[FEATURE] More example scenes


sampaiodias:
**Is your feature request related to a problem? Please describe.**
The concepts of Unity Atoms and the ScriptableObject workflow in general are a little complex to understand for some developers.

**Describe the solution you'd like**
Creation of more examples to guide people through the features of Unity Atoms and suggested workflows.

**Describe alternatives you've considered**
YouTube videos, improving the documentation website. Both are valid, this is just a more "hands-on" approach.

**Additional context**
None.

soraphis:
yeah, i already thought about a larger example. but currently I have not that much time.

I thought about a small diablo-like example "game".
you've got your hero, an action bar with health, mana, skills.
some monsters with floating health bars. a larger healthbar showing the hovered monster.

AdamRamberg:
Closing due to inactivity. There is already a new more complex example scene in the project.

PR #80, opened 2019-11-10
New example scene: Score

sampaiodias:
Simple example that uses Variables, Events and Event Listeners. This PR is related to #79


sampaiodias:
I accidentally added two other commits to the pull request. Is there an option to remove those from the PR?

PR #81, opened 2019-11-10
New example scene: Score

sampaiodias:
Simple example that uses Variables, Events and Event Listeners. Audio files are available [here ](https://opengameart.org/content/100-cc0-sfx) and are CC0 licensed.

This PR is related to #79


AdamRamberg:
Thanks for submitting this PR! 

Tried to open this locally, but the scene seems to be broken. Not sure if I'm doing something wrong here, but could you take a look at it again and see if you don't have any local changes not pushed? 

I'm also getting errors for not having platform `BJM` (probably due to that you have a platform installed that I don't). We should not exclude platforms explicitly. Check out other `.asmdef` files in the project. 

Would also like to know more about the Conditions that you have committed. What is the purpose of these and how do you use them and work with them? Is it something you would like to be added to the library or just to the example?

AdamRamberg:
@sampaiodias Are you still working on this? Thinking about closing this PR if it's not active.

AdamRamberg:
Closing PR due to inactivity.

PR #84, opened 2019-11-11
changed AddComponentMenu names for Listeners/Hooks

IceTrooper:
This PR is related to #83

PR #86, opened 2019-11-11
#84 parameterless atom listener

IceTrooper:
This PR is related to #84 Just last commit

### DIFF
--- a/Packages/Core/Runtime/Actions/AtomAction.cs
+++ b/Packages/Core/Runtime/Actions/AtomAction.cs
@@ -6,7 +6,6 @@ namespace UnityAtoms
     /// <summary>
     /// Base abstract class for Actions. Inherits from `BaseAtom`.
     /// </summary>
-    [EditorIcon("atom-icon-purple")]
     public abstract class AtomAction : BaseAtom
     {
         /// <summary>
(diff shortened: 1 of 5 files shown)


soraphis:
I cherrypicked the one commit for this PR and force pushed it onto the the branch (based on current canary). pls verify that everything is alright. 
thanks

AdamRamberg:
Thanks for the PR! 

Could you do the following changes: 
In `AtomListener.cs` and in the method(s) `OnEventRaised`, please use this syntax `if (_unityEventResponse != null) { _unityEventResponse.Invoke(first, second); }` instead of the null conditional. See https://github.com/dotnet/csharplang/issues/2020
Remove the `[EditorIcon("atom-icon-purple")]` attribute in  `AtomAction`. This was actually never needed in the first place in this file. Needs instead to exist in the final implementation of an AtomAction. 

After the above is fixed I will merge the PR.

soraphis:
> See dotnet/csharplang#2020

UnityEvent inherits from UnityventBase (which inherits from nothing).  it does not inherit from UnityEngine.Object and should not be affected from the null-propagation problematic.

just saying. I actually don't care whether its an if or the propagation syntax.

AdamRamberg:
> UnityEvent inherits from UnityventBase (which inherits from nothing). it does not inherit from UnityEngine.Object and should not be affected from the null-propagation problematic.
> 
> just saying. I actually don't care whether its an if or the propagation syntax.

Aha, that makes sense! Disregard that particular change then :)

IceTrooper:
Done. I thought we need it because you got EditorIcon in every AtomAction\<T\> and etc. Anyway it's done.

Issue #27, opened 2019-04-05
Variable transformers, eg. Clamp Value


AdamRamberg:
For Variables, set a ShouldUpdate "hook" that you query before you update the value. Eg: 
`
        public bool SetValue(T value)
        {
if (!AreEqual(this.value, value) && (ShouldUpdate == null  || (ShouldUpdate != null && ShouldUpdate(T newValue, T currentValue, T oldValue))))
            {
                 // Set value
            }
        }
`
This makes it possible to add your custom logic to your variables, for example "clamp value" - to clamp the value between a min and a max. In this feature it is also included to create the "clamp value" hook.

soraphis:
instead of "hooks" (which sounds to me like 'event system'), it would be possible to make variables contain a list of "value preprocessors".

when a variable is set to a new value, the new value is modified by those preprocessors first.

an example implementation to demonstrate this:

```Csharp
public abstract class ValuePreProcessor<T>{
    public abstract T Process(T value);
}

public abstract class ScriptableVariable<T> : ScriptableObject{

    private T value;
    public T Value{
        get{ return value; }
        set{
            var val = value;
            for(int i = 0; i < preProcessors.Length; ++i){
                val = preProcessors.Process(val);
            }
            this.value = val;
        }
    }

    public ValuePreProcessor<T>[] preProcessors;

}

public class FloatVariable : ScriptableVariable<float>{}

public class ClampPreProcess : ValuePreProcessor<float>{
    public FloatReference minValue;
    public FloatReference maxValue;

    public override float Process(float value){
        return Mathf.Clamp(value, minValue.Value, maxValue.Value);
    }
}
```

jeffcampbellmakesgames:
I think a simpler implementation of this for numeric values would be to add new fields on each closed numeric scriptable variable type  for asserting that the value should be constrained, the min/max of the range, and then `SetValue` would use these internally when setting the value to clamp it to the desired range. It would require a small amount of additional code per type, but this would be mostly a copy-paste with maybe some slight variation. Adding another OOP level to `SetValue` for processing simple value types seems like overkill.

AdamRamberg:
I really like @soraphis solution - it's flexible and extendible. ValuePreProcessor should be an array of GameFunctions though like this: 
`
      public abstract class ScriptableObjectVariable<T, E1, E2, GF> : ScriptableObjectVariableBase<T>,
        IWithOldValue<T>,
        ISerializationCallbackReceiver
[...]

PR #93, opened 2019-11-19
Added pre change transformers to Variables + Clamp Int / Float

AdamRamberg:
This PR addresses #27. Things added in this PR: 
- Added a list of pre change transformers to AtomVariable
- Added docs
- Added two AtomFunctions - ClampInt and ClampFloat
- Added custom editors for ClampInt and ClampFloat
- AtomFunction<T, T> is now part of the generator / regenerate all atoms
- Regenerated all the Atoms

### DIFF
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -57,16 +57,36 @@ namespace UnityAtoms
         /// <summary>
         /// When setting the value of a Variable the new value will be piped through all the pre change transformers, which allows you to create custom logic and restriction on for example what values can be set for this Variable.
         /// </summary>
+        /// <value>Get the list of pre change transformers.</value>
+        public List<F> PreChangeTransformers
+        {
+            get => _preChangeTransformers;
+            set
+            {
+                if (value == null)
+                {
+                    _preChangeTransformers.Clear();
+                }
+                else
+                {
+                    _preChangeTransformers = value;
+                }
+            }
+        }
+
         [SerializeField]
-        public List<F> PreChangeTransformers;
+        private List<F> _preChangeTransformers = new List<F>();
 
         protected abstract bool AreEqual(T first, T second);
 
+        private void OnValidate()
+        {
+            _initialValue = RunPreChangeTransformers(_initialValue);
+            _value = RunPreChangeTransformers(_value);
+        }
+
         private void OnEnable()
         {
-            if (PreChangeTransformers == null) { PreChangeTransformers = new List<F>(); }
-            _initialValue = RunPreChangeTransformers(_initialValue);
-
             _oldValue = _initialValue;
             _value = _initialValue;
 
@@ -159,18 +179,21 @@ namespace UnityAtoms
 
         private T RunPreChangeTransformers(T value)
         {
-            var preProcessedValue = value;
-            if (PreChangeTransformers != null && PreChangeTransformers.Count > 0)
+            if (_preChangeTransformers.Count <= 0)
             {
-                for (var i = 0; i < PreChangeTransformers.Count; ++i)
+                return value;
+            }
+
+            var preProcessedValue = value;
+            for (var i = 0; i < _preChangeTransformers.Count; ++i)
+            {
+                var Transformer = _preChangeTransformers[i];
+                if (Transformer != null)
                 {
-                    var Transformer = PreChangeTransformers[i];
-                    if (Transformer != null)
-                    {
-                        preProcessedValue = Transformer.Call(preProcessedValue);
-                    }
+                    preProcessedValue = Transformer.Call(preProcessedValue);
                 }
             }
+
 
             return preProcessedValue;
         }
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using UnityEngine;
 
 namespace UnityAtoms
@@ -9,10 +10,12 @@ namespace UnityAtoms
     /// <typeparam name="T">The Variable value type.</typeparam>
     /// <typeparam name="E1">Event of type `AtomEvent&lt;T&gt;`.</typeparam>
     /// <typeparam name="E2">Event of type `AtomEvent&lt;T, T&gt;`.</typeparam>
+    /// <typeparam name="F">Function of type `FunctionEvent&lt;T, T&gt;`.</typeparam>
     [EditorIcon("atom-icon-lush")]
-    public abstract class AtomVariable<T, E1, E2> : AtomBaseVariable<T>
+    public abstract class AtomVariable<T, E1, E2, F> : AtomBaseVariable<T>
         where E1 : AtomEvent<T>
         where E2 : AtomEvent<T, T>
+        where F : AtomFunction<T, T>
     {
         /// <summary>
         /// The Variable value as a property.
@@ -51,10 +54,19 @@ namespace UnityAtoms
         /// </summary>
         public E2 ChangedWithHistory;
 
+        /// <summary>
+        /// When setting the value of a Variable the new value will be piped through all the pre change transformers, which allows you to create custom logic and restriction on for example what values can be set for this Variable.
+        /// </summary>
+        [SerializeField]
+        public List<F> PreChangeTransformers;
+
         protected abstract bool AreEqual(T first, T second);
 
         private void OnEnable()
         {
+            if (PreChangeTransformers == null) { PreChangeTransformers = new List<F>(); }
+            _initialValue = RunPreChangeTransformers(_initialValue);
+
             _oldValue = _initialValue;
             _value = _initialValue;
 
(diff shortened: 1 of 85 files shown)


soraphis:
reading over the code:

- ClampFloatEditor and ClampIntEditor are (almost) identical this could be changed to a single "ClampEditor"

    [CustomEditor(typeof(ClampInt))]
    [CustomEditor(typeof(ClampLong))]
    [CustomEditor(typeof(ClampFloat))]
    [CustomEditor(typeof(ClampDouble))]
    public class ClampEditor : UnityEditor.Editor
    {
     // ...


they could implement an interface defining the IsValid call, and 'target' could be casted to the interface.

will check out the commit within the next day, and look into it in more detail

Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -142,5 +156,23 @@ public IObservable<T> ObserveChange()
...
+
+        private T RunPreChangeTransformers(T value)
+        {
+            var preProcessedValue = value;
+            if (PreChangeTransformers != null && PreChangeTransformers.Count > 0)
+            {
+                for (var i = 0; i < PreChangeTransformers.Count; ++i)
+                {
+                    var Transformer = PreChangeTransformers[i];
+                    if (Transformer != null)
```
soraphis: filter all "null"s out of the List in OnEnable() to reduce null checks against UnityEngine.Objects (which ScriptableObjects are)?

Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -51,10 +54,19 @@ public abstract class AtomVariable<T, E1, E2> : AtomBaseVariable<T>
...
+        /// When setting the value of a Variable the new value will be piped through all the pre change transformers, which allows you to create custom logic and restriction on for example what values can be set for this Variable.
+        /// </summary>
+        [SerializeField]
+        public List<F> PreChangeTransformers;
+
         protected abstract bool AreEqual(T first, T second);
 
         private void OnEnable()
         {
+            if (PreChangeTransformers == null) { PreChangeTransformers = new List<F>(); }
```
soraphis: this is always false, because its filled by serialization. also: just initialize a new list at the declaration, to be sure it exists

Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -51,10 +54,19 @@ public abstract class AtomVariable<T, E1, E2> : AtomBaseVariable<T>
...
+            _initialValue = RunPreChangeTransformers(_initialValue);
```
soraphis: should be in OnValidate()

soraphis:
**will check out the commit within the next day, and look into it in more detail**

(have not done this yet, will edit this post when done)

Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -51,10 +54,19 @@ public abstract class AtomVariable<T, E1, E2> : AtomBaseVariable<T>
         /// </summary>
         public E2 ChangedWithHistory;
 
+        /// <summary>
+        /// When setting the value of a Variable the new value will be piped through all the pre change transformers, which allows you to create custom logic and restriction on for example what values can be set for this Variable.
+        /// </summary>
+        [SerializeField]
+        public List<F> PreChangeTransformers;
```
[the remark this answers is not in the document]
AdamRamberg: Fixed in the latest commit.

Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -142,5 +156,23 @@ public IObservable<T> ObserveChange()
             );
         }
         #endregion // Observable
+
+        private T RunPreChangeTransformers(T value)
+        {
+            var preProcessedValue = value;
+            if (PreChangeTransformers != null && PreChangeTransformers.Count > 0)
```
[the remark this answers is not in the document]
AdamRamberg: Fixed in the latest commit.

Packages/Core/Runtime/Variables/AtomVariable.cs
AdamRamberg: Don't think we should skip the null check since it is possible to add change transformers at runtime (you could potentially add null).

Packages/Core/Runtime/Variables/AtomVariable.cs
AdamRamberg: Fixed in the latest commit.

Packages/Core/Runtime/Variables/AtomVariable.cs
AdamRamberg: Fixed in the latest commit.

AdamRamberg:
**Fixes in the latest commit:** 
- Created a base class for the editor for ClampFloat and ClampInt. Apparently it is not allowed to have multiple CustomEditor attributes defined for the same class. To implement this I also created an interface called `IIsValid` (for casting purposes in the editor). 
- Initialize `PreChangeTransformers ` at declaration and removed initialization `OnEnable`
- Moved call to `RunPreChangeTransformers ` from `OnEnable` to `OnValidate`. Running it 2 times, once for initialValue and once for value since value can be changed at runtime via editor. 
- Made the variable PreChangeTransformers private and created a property instead in order to make it impossible to overwrite the list with `null` from the outside.
- Moved `ClampInt` and `ClampFloat` in the `CreateAssetMenu` for `Unity Atoms/Functions/Transformers`

AdamRamberg:
> From this few changes: its super tedious to work with it. you need to create an additional asset for every time you use the clamp function. this should be easier.
> 
> Maybe its possible to add an Attribute to a method, scan the assembly for all matching attributes and show a searchable list add the Preprocessor list?
> 
> i think in its current state its no fun to work with it.

I get what you are saying. Still think it is neat to have functions as ScriptableObjects, but it would be nice to make the experience working with them better. Will look into the Attribute option and come back. Let me know if you come up with some other solution that we might have missed.

soraphis:
>  Apparently it is not allowed to have multiple CustomEditor attributes defined for the same class.

Oo thats strange. 🤔 is this a newer change? i thought this was possible.
but you could use the BaseClass, and tell it to use it as ["fallback = true"](https://docs.unity3d.com/ScriptReference/CustomEditor-isFallback.html) IIRC.

> Still think it is neat to have functions as ScriptableObjects

maybe a function. but with this solution, we have each individual function call a object. dozenz of Clamp's. Clamp health, Clamp mana, Max Gold, ...

Maybe using the decorator pattern?
Lets say there is a `FloatVariable : AtomVariable<float, ...>` and there are 

```
    [Decorates(typeof(FloatVariable))] // maybe not needed
    public class MinFloatVariable : FloatVariable{    public MinFloatVariable(FloatVariable v) // ....  

    [Decorates(typeof(FloatVariable))] // maybe not needed
    public class MaxFloatVariable : FloatVariable{    public MaxFloatVariable(FloatVariable v) // ....  
```

Min and Max use the decorator pattern to add logic. in the Variable editor you'd have a dropdown list at the top/bottom which lets you select decorations (can be found with the help of the Attribute). when chosen, this Variable is replaced with a new one (of the decorating type, same name) containing the old one. 

the editor of a decorating type, should show the additional fields and then render the editor of the containing type.

I'll test this, this evening.

AdamRamberg:
> checked out this branch. went to demo scene and created a new "Function > Clamp Int" named it ClampHealth.
> 
> started to throw exceptions:

Is this a problem after the latest commit? Can't reproduce it. Maybe you are running a different version of Unity? It's weird because the error message doesn't either seem to be related to the change.

AdamRamberg:
@soraphis Regarding your suggestion above, it will unfortunately not work if we don't add custom serialization to Unity Atoms. The problem is Unity's serializer won't serialize the dynamically added decorators. So in my mind this route is not viable. 

I'm thinking that even though it is a little bit clunky DX / UX wise that this is the Unity Atoms way of solving this feature. I propose to add it, but also work on some separate features to make the DX / UX more pleasant. What suggest to add besides this PR is: 
- A custom drawer for the list of pre change transformers (but could be used for other lists in Unity Atoms). The drawer will make it easier to add / remove items from the list, adding buttons for that (the same way as the inspector for event handlers like for button click works in vanilla Unity).
- Add a "inspector context". This means that if you have an atom in a list, like the pre change transformers, you can click on "go to atom". When you are viewing the inspector for that Atom you will have a back button getting you back to the last Atom shown in the inspector. This would enable a much smoother workflow with something like the pre change transformers. 

@soraphis What are your thoughts? Anyone else got some feedback on this?

soraphis:
>Regarding your suggestion above, it will unfortunately not work if we don't add custom serialization to Unity Atoms.

No custom serialization needed. but probably a to complex set of features for this.

> this is the Unity Atoms way of solving this feature. 

I'm okay with this

> Add a "inspector context".

I like this.

AdamRamberg:
> No custom serialization needed. but probably a to complex set of features for this.

I'm curious, how would you go about it without implementing some custom serialization?

soraphis:
> I'm curious, how would you go about it without implementing some custom serialization?

in my code snipped above, ignore the constructor call, it should be just a property.

fetching decorator-classes from the compiled assembly and show a list of their names.
those decorator-classes inherit the <Type>Variable class, so they are scriptable objects, just instantiate a new one, when something in the list is selected.
move the current asset to be a sub-asset of the new created and hide it.

make the decorator reference it in a hidden but serialized variable.

it's all just asset referencing. but it's a lot of hassle to manage all that without introducing bugs.

AdamRamberg:
Ah didn't know about sub assets to be honest. Great to learn something new! This would just work for one decorator per Variable though right? I get that it technically would work with multiple decorators, but the editor will just be shown for the latest one?

PR #98, opened 2020-01-16
Unity v2018.4 Support

fakegood:
- Fixed MonoHooks asmdef reference to Core asmdef was broken.
- Fixed dependancy issue for Package Manager.
- Removed Generator function in version 2018.4.

### DIFF
--- a/Packages/Core/Editor/Drawers/AtomDrawer.cs
+++ b/Packages/Core/Editor/Drawers/AtomDrawer.cs
@@ -1,4 +1,3 @@
-#if UNITY_2019_1_OR_NEWER
 using System.Collections.Generic;
 using UnityEditor;
 using UnityEngine;
@@ -148,4 +147,3 @@ namespace UnityAtoms.Editor
         }
     }
 }
-#endif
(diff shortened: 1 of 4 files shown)


Packages/Core/Editor/Drawers/AtomDrawer.cs
```diff
@@ -1,4 +1,3 @@
-#if UNITY_2019_1_OR_NEWER
```
soraphis: should be 2018_4_or_newer
instead of removing, or?

fakegood: updated to have UNITY_2018.4_OR_NEWER and also updated the summary comment.

fakegood:
> > Removed Generator function in version 2018.4.
> 
> why?

because RegenerateAllAtoms will not work properly. GeneratorEditor will need a re-write to support 2018.4 also. I will submit another PR that would have all of the Generator support. Or you prefer it to be in this PR?

AdamRamberg:
Thanks for the PR! ❤️ It looks good! However, could you please make the PR against the canary branch instead? (they are on the same commit so should be easy) After that I will merge this one. 

You can add an issue + create a new PR for making RegenerateAllAtoms + GeneratorEditor support 2018.4.

soraphis:
@AdamRamberg i changed the base to canary

Issue #99, opened 2020-01-20
[FEATURE] published on OpenUPM


favoyang:
**Is your feature request related to a problem? Please describe.**

Hi @AdamRamberg,

Thanks for making the great UPM package.

I'm the creator of [OpenUPM](https://openupm.com), a managed UPM registry with automatic build pipelines. The services monitor git tags and publish new package releases regularly. You can read the [announcement ](https://medium.com/openupm/openupm-beta-is-now-available-a6665ff60c71) for more information.

I have included your packages into our build pipelines,
https://openupm.com/packages/com.mambojambostudios.unity-atoms-core/
https://openupm.com/packages/com.mambojambostudios.unity-atoms-mobile/
https://openupm.com/packages/com.mambojambostudios.unity-atoms-mono-hooks/
https://openupm.com/packages/com.mambojambostudios.unity-atoms-scene-mgmt/
https://openupm.com/packages/com.mambojambostudios.unity-atoms-tags/
https://openupm.com/packages/com.mambojambostudios.unity-atoms-ui/

I hope you like the idea to have an additional but UPM exclusive registry to host your packages.

You don't need to do anything extra. Your monorepo repository is well organized. I'll prepare a PR for the documentation update and badges, please consider to merge if looks okay.

favoyang:
Added PR https://github.com/AdamRamberg/unity-atoms/pull/100 for the documentation update and badges

AdamRamberg:
@favoyang Thanks a lot! OpenUPM looks really great! Your PR was just merged to the canary branch. The instructions will included and added to master when we release the next version of Unity Atoms.

PR #100, opened 2020-01-20
docs: add installation options and badges

favoyang:
Refs #99


AdamRamberg:
Hi @favoyang! Your project OpenUPM is really awesome. Thanks for making it <3 Added a commit to add the instructions to install via OpenUPM to the website as well.

Issue #101, opened 2020-01-21
[Bug] Cant remove tag in Runtime


lucasrferreira:
The following code doesnt work.

```csharp
AtomTags atomTags = GetComponent<AtomTags>();
atomTags.RemoveTag(TagToRemove.Value);
```

Tested in 2.0.0 and 2.1.0

[EDIT]
I think I found the error. On "RemoveTag" function, is tested if tag is in Dict, but it return if true and not if false.

```csharp
        /// <summary>
        /// Remove a tag from this `GameObject`.
        /// </summary>
        /// <param name="tag">The tag to remove as a `string`</param>
        public void RemoveTag(string tag)
        {
            if (tag == null) return;
            if (_sortedTags.ContainsKey(tag)) return;  \\ <<<<<<---- HERE
            _sortedTags.Remove(tag);

            Tags = new ReadOnlyList<StringConstant>(_sortedTags.Values);

            // Update static accessors:
            if (!TaggedGameObjects.ContainsKey(tag)) return; // this should never happen
            TaggedGameObjects[tag].Remove(this.gameObject);
        }
```

soraphis:
thank you for your submission.

i think you're right. 
the few tests I wrote don't cover removing (I wrote them mostly to measure GC allocations)

could you create a pull request?
thank you.

lucasrferreira:
Created PR: https://github.com/AdamRamberg/unity-atoms/pull/102

I had fixed the tests to.

soraphis:
> I have a project current in development using that module. Will you release a new version if that fix soon?

@AdamRamberg you're in a better position to answer to that

AdamRamberg:
Hi @lucasrferreira! Thanks for the issue and the PR. Just merged it into canary. I will try to release a new version of Unity Atoms at the latest this weekend, but hopefully I will get the time tonight to do it.

AdamRamberg:
@lucasrferreira v2.1.1 just got released and includes the changes in this PR :)

PR #102, opened 2020-01-22
Fix Remove Tag from AtomTags and Fix Remove Tag Test

lucasrferreira:
ref: #101

Fixed Remove tag and and Fixed tests to cover that function.


AdamRamberg:
Thanks for this PR! Looks good! Merging to canary.

PR #106, opened 2020-02-16
- Reactivate Generator function in Unity 2018.4.

fakegood:
Since Generator is one of the key features of Unity Atoms, this PR is to re-activate this feature. And since 2018.4 is LTS version I think it should be supported as well.


Packages/Core/Editor/Editors/AtomEventEditor.cs
```diff
@@ -1,6 +1,10 @@
-#if UNITY_2019_1_OR_NEWER
+#if UNITY_2018_4_OR_NEWER
 using UnityEngine;
+#if UNITY_2018_4_OR_NEWER
```
AdamRamberg: The UNITY_2019_1_OR_NEWER condition should be before UNITY_2018_4_OR_NEWER, the second if-statement will never be used when running Unity 2019 since the UNITY_2018_4_OR_NEWER flag will also be true in this case.

Packages/Core/Editor/Editors/AtomEventEditor.cs
```diff
@@ -12,6 +16,9 @@ namespace UnityAtoms.Editor
     public abstract class AtomEventEditor<T, E> : UnityEditor.Editor
         where E : AtomEvent<T>
     {
+#if UNITY_2018_4_OR_NEWER
```
AdamRamberg: Same as above. The UNITY_2019_1_OR_NEWER condition should be before UNITY_2018_4_OR_NEWER, the second if-statement will never be used when running Unity 2019 since the UNITY_2018_4_OR_NEWER flag will also be true in this case.

Examples/Packages/manifest.json
```diff
@@ -9,15 +9,8 @@
     "com.mambojambostudios.unity-atoms-tags": "file:../../Packages/Tags",
     "com.mambojambostudios.unity-atoms-ui": "file:../../Packages/UI",
     "com.neuecc.unirx": "https://github.com/starikcetin/UniRx.git#7.1.0-upm",
-    "com.unity.2d.sprite": "1.0.0",
```
AdamRamberg: Even though we support 2018.4, the Example project itself should run Unity 2019, and some of these packages are needed in that version. Please revert this file.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -58,7 +58,11 @@ public void Generate(string type, string baseWritePath, bool isEquatable, List<A
             var templateSearchPath = Runtime.IsUnityAtomsRepo ?
                 Directory.GetParent(baseWritePath).FullName : // "Packages"
                 Directory.GetParent(Application.dataPath).FullName;
+#if UNITY_2018_4_OR_NEWER
```
AdamRamberg: Same as above. The UNITY_2019_1_OR_NEWER condition should be before UNITY_2018_4_OR_NEWER, the second if-statement will never be used when running Unity 2019 since the UNITY_2018_4_OR_NEWER flag will also be true in this case.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -70,20 +74,25 @@ public void Generate(string type, string baseWritePath, bool isEquatable, List<A
 
             foreach (var templatePath in templatePaths)
             {
-                var templateNameStartIndex = templatePath.LastIndexOf(Path.DirectorySeparatorChar) + 1;
+#if UNITY_2018_4_OR_NEWER
```
AdamRamberg: Same as above + add UNITY_2019_1_OR_NEWER condition. The UNITY_2019_1_OR_NEWER condition should be before UNITY_2018_4_OR_NEWER, the second if-statement will never be used when running Unity 2019 since the UNITY_2018_4_OR_NEWER flag will also be true in this case.

Packages/Core/Editor/Generator/GeneratorEditor.cs
```diff
@@ -1,10 +1,16 @@
-#if UNITY_2019_1_OR_NEWER
+#if UNITY_2018_4_OR_NEWER
 using System;
 using System.Linq;
 using System.Collections.Generic;
 using UnityEditor;
 using UnityEngine;
+#if UNITY_2018_4_OR_NEWER
```
AdamRamberg: Same as above. The UNITY_2019_1_OR_NEWER condition should be before UNITY_2018_4_OR_NEWER, the second if-statement will never be used when running Unity 2019 since the UNITY_2018_4_OR_NEWER flag will also be true in this case.

Packages/Core/Editor/Generator/GeneratorEditor.cs
```diff
@@ -147,7 +153,11 @@ private void OnEnable()
             };
             generator = generator == null ? new Generator() : generator;
 
+#if UNITY_2018_4_OR_NEWER
```
AdamRamberg: Same as above. The UNITY_2019_1_OR_NEWER condition should be before UNITY_2018_4_OR_NEWER, the second if-statement will never be used when running Unity 2019 since the UNITY_2018_4_OR_NEWER flag will also be true in this case.

Examples/Packages/manifest.json
```diff
@@ -59,4 +59,4 @@
       "revision": "7.1.0-upm"
     }
   }
-}
+}
```
AdamRamberg: Please revert this change

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -58,7 +58,11 @@ public void Generate(string type, string baseWritePath, bool isEquatable, List<A
             var templateSearchPath = Runtime.IsUnityAtomsRepo ?
                 Directory.GetParent(baseWritePath).FullName : // "Packages"
                 Directory.GetParent(Application.dataPath).FullName;
+#if UNITY_2019_1_OR_NEWER
```
AdamRamberg: This check is not necessary. You should always be able to do: 
`var templatePaths = Directory.GetFiles(templateSearchPath, "UA_Template*.txt", SearchOption.AllDirectories);`

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -70,20 +74,25 @@ public void Generate(string type, string baseWritePath, bool isEquatable, List<A
 
             foreach (var templatePath in templatePaths)
             {
-                var templateNameStartIndex = templatePath.LastIndexOf(Path.DirectorySeparatorChar) + 1;
+#if UNITY_2019_1_OR_NEWER
```
AdamRamberg: All the changes from this line and below is not necessary, it should work with just using `templatePath`.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -70,20 +74,25 @@ public void Generate(string type, string baseWritePath, bool isEquatable, List<A
...
+                var localTemplatePath = AssetDatabase.GUIDToAssetPath(templatePath);
+#endif
+                var templateNameStartIndex = localTemplatePath.LastIndexOf(Path.DirectorySeparatorChar) + 1;
                 var fileExtLength = 4;
-                var templateName = templatePath.Substring(templateNameStartIndex, templatePath.Length - templateNameStartIndex - fileExtLength);
+                var templateName = localTemplatePath.Substring(templateNameStartIndex, localTemplatePath.Length - templateNameStartIndex - fileExtLength);
                 var lastIndexOfDoubleUnderscore = templateName.LastIndexOf("__");
                 var atomType = templateName.Substring(lastIndexOfDoubleUnderscore + 2);
-                var capitalizedAtomType = Capitalize(atomType);
+                var capitalizedAtomType = Capitalize(atomType); Debug.Log(localTemplatePath + " : " + templateName);
```
AdamRamberg: Don't forget to remove the `Debug.Log`

fakegood:
Updated based on comments.

Added path to Core Package because the template search path should not be based on a variable path input by user.

When Example project is opened, changed Generator path to "Assets/Atoms" will cause search path to search in "Assets/Atoms" instead of "../Packages/Core", which will not be able to search for any templates.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -56,8 +56,9 @@ public void Generate(string type, string baseWritePath, bool isEquatable, List<A
 
             // Recursively search for template files. TODO: Is there a better way to find and load templates?
             var templateSearchPath = Runtime.IsUnityAtomsRepo ?
-                Directory.GetParent(baseWritePath).FullName : // "Packages"
+                Directory.GetParent(Runtime.UnityAtomsCorePath).FullName : // "Packages"
```
AdamRamberg: I would prefer if you do it like this instead:
`Directory.GetParent(Application.dataPath).Parent.FullName`
In my mind it's better if we can compute the path at runtime instead of relying on constants.

Packages/Core/Runtime/Utils/Runtime.cs
```diff
@@ -26,5 +26,14 @@ public static bool IsUnityAtomsRepo
         {
             get => System.Environment.CurrentDirectory.Contains("unity-atoms/Examples");
         }
+
+        /// <summary>
+        /// Returns the path to Unity Atoms Core.
+        /// Mainly used to search for Templates.
+        /// </summary>
+        public static string UnityAtomsCorePath
```
AdamRamberg: Please remove constant. See my first comment.

fakegood:
Updated based on your comments. :)

Packages/Core/Editor/Drawers/VariableDrawer.cs
```diff
@@ -14,15 +14,29 @@ public override void OnGUI(Rect position, SerializedProperty property, GUIConten
                 return;
             }
 
+            label = EditorGUI.BeginProperty(position, label, property);
```
AdamRamberg: Sorry, but need you to revert these changes before I merge this PR (the complete file).

fakegood:
I've updated the pull request.

PR #109, opened 2020-02-16
Fix indentation bug in variable drawer

### LINKED ISSUE #104: [BUG] Variables in lists are not drawn correctly
**Describe the bug**
When using variables in lists, _e.g. List\<FloatVariable\>_, the items are not drawn correctly once a variable has been assigned to an element in the list. Some items drawn by AtomDrawer & VariableDrawer are overlapping (see screenshot)

**To Reproduce**
1. Create a List of some Atom Variable in a MonoBehaviour
```
public class AtomDRawerIssue : MonoBehaviour
{
  // ...
  public List<FloatVariable> vars;
  // ...
}
```
2. Assign a FloatVariable to an element in the list in the inspector

**Expected behavior**
Assigned elements should be drawn without overlapping.

**Screenshots**
The first item in the list has been assigned and shows the overlapping issue:
Screenshot 2020-02-10 at 22 02 39

**Environment**
 - OS: OSX 10.14.5 (Mojave)
 - Unity Version 2019.3.0f3
 - Version of Unity Atoms v2.1.0

**Additional context**
I've tested with a few variable types but not all. Line 19 in VariableDrawer.cs seems to be part of the problem, the width is calculated using depth, and in the case of the list depth is not 0. Removing the depth condition on that line fixes the issue for Lists, but breaks drawing when using References.

AdamRamberg:
This PR fixes #104. The solution is using a smilar strategy as how the AtomReferenceDrawer is solving laying out its GUI horizontally.

### DIFF
--- a/Packages/Core/Editor/Drawers/VariableDrawer.cs
+++ b/Packages/Core/Editor/Drawers/VariableDrawer.cs
@@ -14,15 +14,29 @@ namespace UnityAtoms.Editor
                 return;
             }
 
+            label = EditorGUI.BeginProperty(position, label, property);
+            position = EditorGUI.PrefixLabel(position, label);
+
             var inner = new SerializedObject(property.objectReferenceValue);
             var valueProp = inner.FindProperty("_value");
-            var width = GetPreviewSpace(valueProp.type) + (property.depth == 0 ? EditorGUIUtility.labelWidth : 0);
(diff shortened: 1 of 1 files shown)


AdamRamberg:
Such a small change that seems to work so I will merge this one to canary.

Issue #66, opened 2019-10-20
Add Variables directly to Event Listeners


mutmedia:
It's a bit annoying to have to create events all the time to be able to use the event listeners, so sometimes it would be nice to be able to set either a Variable or an Event to the listener (making the event be used in case it exists, or created on load in case it's not).

AdamRamberg:
Thanks for submitting the issue. I really like this idea! However, I don't think auto creating an event if it doesn't exist is a good idea, since it might not preferable in all cases. 

**My take on this feature:** 

**Is your feature request related to a problem? Please describe.**
Better DX. See comment above by @mutmedia

_This is only applicable for Listeners listening to one type._ 

**Describe the solution you'd like**
Create a new member property in the `AtomListener` class for a Variable of the same type as the current `Event`. Make a PropertyDrawer for the Event / Variable pair OR make a custom Editor for the `AtomListener`. The Editor / PropertyDrawer should be similar to the PropertyDrawer for References, eg. choose Event or Variable in a "Popup" (`Event` by default) and then corresponding PropertyField. 

**Describe alternatives you've considered**
Not creating a custom PropertyDrawer. However, I think it's preferable to do so. 

**Additional note**
Use `UnityEngine.UIElements` when developing the Editor / PropertyDrawer. All new Editor development should be done using the new UIElements since that is the direction that Unity is going. 

I would really appreciate a PR for this. Read the [contribution guidelines](https://github.com/AdamRamberg/unity-atoms/blob/canary/CONTRIBUTING.md) before starting.

AdamRamberg:
Fixed in #111 - You can now use AtomEventReference for this.

Issue #53, opened 2019-08-22
Add an AtomInstancer component


IceTrooper:
![obraz](https://user-images.githubusercontent.com/51023959/63531648-42f15300-c509-11e9-9750-878ffcc9bd47.png)
http://www.roboryantron.com/2017/10/unite-2017-game-architecture-with.html?showComment=1518505684655#c499355616768136746

What do you think about that? Do we need that? It's based on Unite 2017 talk.

AdamRamberg:
@IceTrooper This is pretty interesting. 

Another way to go about this is to create new instances of the SO in `OnEnable`without an instancer. The only functional difference really would be that you can't reuse the Variable as a template in several places, but not sure there is a real use case for this. However, using an instancer will make a script using it more game designer friendly. I'll say that we should implement this into atoms. If you are up to it would be much appreciated with a PR.

AdamRamberg:
**My thoughts on this**
I would like the API to look like like this: 
```
var myIntVariable = GetComponent<AtomInstancer>().Get<IntVariable>(key: "my_key");
```
**Behaviour:**
The AtomInstancer (`MonoBehaviour`) will have a Dictionary / Map of Atoms, containing a Key (string) and a Value (an Atom of any kind). The `Get<T>(string key = null)` method will first try to locate an Atom of type T (T will have the constrain `where T : BaseAtom`). If `key` is omitted it will just pick the first it finds of type `T`, otherwise it does a match on the `key` value. If it finds a match it returns the Atom, otherwise it return `null`.

I would be really appreciate someone making a PR on this. Should not be too much work to get a first version up. Make sure to read the [guidelines](https://github.com/AdamRamberg/unity-atoms/blob/canary/CONTRIBUTING.md) before starting + assign yourself (or ask me) to the issue.

AdamRamberg:
Added in #110

Issue #71, opened 2019-10-31
[FEATURE][PROPOSAL] AtomCollections


soraphis:
**Is your feature request related to a problem? Please describe.**
The want to group many UnityAtomVariables together, for more structure.

(Possible other use cases, as ViewModel for a View in an MVVM like pattern (UI stuff).

**Describe the solution you'd like**
The possibility to create a "AtomVariableCollection" asset where you can drag-and-drop your atoms as sub-assets into.

![image](https://user-images.githubusercontent.com/3066582/67988216-97aef980-fc2f-11e9-9d72-b327d90a6c24.png)

**This is an Proposal**
Discussion is encouraged. Possible Use-Cases, Features to include.

sampaiodias:
This would be super useful if users could put names on the fields instead of an array. We could access the variables by using collection.Get("Character") or something like that instead of collections.atoms[1]

AdamRamberg:
Really nice feature! 👌 

@sampaiodias I agree that the getting Atoms by a specified key would be neat. I assume you mean that `Get` would return the value of the Variable right? Furthermore, I think this could be pretty tricky to implement though without boxing and / or using reflection due to that we don't know if its a reference or a value type when casting (maybe my knowledge here in this eara is too limited though). That said, would love a solution for this. 

@sampaiodias JSON serialization / deserialization also sounds neat!

sampaiodias:
Maybe it would help if the AtomCollection had an array of AtomCollectionItem instead of AtomVariable or something like that.

```csharp
[System.Serializable]
public class AtomCollectionItem<T> where T : AtomVariable
{
    public string id;
    public T variable;
}
```

soraphis:
**idea**:
Maybe the Collection should be a listener, automatically listen to all Variables change events

and itself send an event whenever something in a collection was changed, so that you can batch multiple changes together and send one event for a bunch of changes.

this could be useful if your listensers would need to ensure that a another variable was set before (which itself could have a listener, ....)

AdamRamberg:
> idea:
> Maybe the Collection should be a listener, automatically listen to all Variables change events
> 
> and itself send an event whenever something in a collection was changed, so that you can batch multiple changes together and send one event for a bunch of changes.
> 
> this could be useful if your listensers would need to ensure that a another variable was set before (which itself could have a listener, ....)

soraphis:
> Another question, how would you identify what Variable was changed on the listener / receiver side?

[...]

PR #110, opened 2020-02-18
Added Variable Instancer, Event Reference, Atom Collection and Atom List (old Atom List renamed to Atom Value List)

AdamRamberg:
This PR adds several features and resolves several reported issues. Here follows the features added in this PR: 

**AtomVariableInstancer**
This adds functionality explained in issue #53. An AtomVariableInstancer is a MonoBehaviour that takes a Variable as a base and creates an in memory cope of it OnEnable. This is particular useful for when creating prefabs that are designed to be instantiated at runtime. Together with AtomEventReference (see below) it solves how to use Atoms for prefabs instantiated at runtime, something that many users have been requesting and asking about. An example of usage would be when creating an enemy prefab you can use an `IntVariableInstancer` that creates an in memory copy of the enemy's health that you then can use in your scripts on the enemy prefab (using References / Event References). 

_Most important things done regarding AtomVariableInstancer:_ 
- Added AtomVariableInstancer as an option to AtomReference.
- Added AtomVariableInstancer to generator.
- Added editor icon for AtomVariableInstancer.

**AtomEventReference**
AtomEventReference is just like a regular Reference, but for Events! You can choose between using Events, Variable (picks the Variable's Changed / ChangedWithHistory event) or Variable Instancer (picks the in memory Variable's Changed / ChangedWithHistory event). This solves issue #66. 

_Most important things done regarding AtomEventReference:_ 
- Added an AtomEventReference class (and AtomEventX2Reference). It’s similar to an AtomReference, but for Events. Let’s you pick between an Event, Variable (will select the Changed event) and a VariableInstancer (see above). 
- Added AtomEventReference and AtomEventX2Reference to generator.
- Added a drawer for AtomEventReference.
- Listeners are now using AtomEventReference instead of AtomEvent.
- Refactoring of VoidHooks since Listeners are now using AtomEventReference.

**Furthermore**
This is a big PR with breaking changes. I think we should test and verify this PR and then (together with some more PR's being developed) merge and release in a new major of Unity Atoms.


AdamRamberg:
Made some new commits adding Atom Collection to this PR, discussed in #71.

**AtomCollection**
- Created an AtomCollection - a collection of Atoms associated with key strings (AtomReferences). 
- Added new editor icon for collections. 
- Created a SerializableDictionary class, which AtomCollection is using.
- Custom property drawer for SerializableDictionary.
- SerializableDictionary supports nested structures meaning that a AtomCollection can have a KVP that is pointing to another AtomCollection.
- AtomCollections have 3 events: Added, Removed, Cleared.
- Added an option to sync an InstanceVariable to collection - adding it to the collection when created (using gameObject’s instance id as key) and removing it from the collection when destroyed.

AdamRamberg:
Made new commits adding AtomList, which is similar to the Collection but a list. Also changed name on AtomList to AtomValueList. 

**AtomList**
- Renamed old AtomList to AtomValueList
- Added AtomList, like Collection, but a list
- Added new icon for AtomList
- Created a AtomBaseVariableList class, which AtomList is using.
- Custom property drawer for AtomBaseVariableList.
- AtomLists have 3 events: Added, Removed, Cleared.
- Added an option to sync an InstanceVariable to list - adding it to the list when created and removing it from the list when destroyed.

Issue #125, opened 2020-04-05
[FEATURE] Raise "Changed With History" event on variable initialization


Dino0040:
**Is your feature request related to a problem? Please describe.**
A health bar that is supposed to react differently when health increases or decreases (which is checked by comparing the value pair of the "Changed With History" event) does not receive an initial value. Only after changing the health value when the game is running the event gets raised.

**Describe the solution you'd like**
Raise the "Changed With History" event on Initialization of a variable.

**Describe alternatives you've considered**
Keep code as is and use both a "Changed" event with and without a history where one changes the health bars display value and the other adds an increase / decrease effect separately.

iruizmar:
This was discussed in Discord since it's not clear if it makes sense, semantically speaking, to call neither Changed nor ChangedWithHistory on OnEnable if the value didn't really changed. 

On the other hand, as @Dino0040 said, it's convenient to make those calls in order to be aware of the Atoms's value on OnEnable anywhere just with the reference to the Event. If not, you would need to hold a reference to the AtomVariable itself too.

So @AdamRamberg proposed to have two boolean values, one for Changed event and the other for ChangedWithHistory event, that indicate if they should be invoked or not on the AtomVariable's OnEnable.

AdamRamberg:
Fixed in #126, which is merged to canary. Closing this issue.

PR #126, opened 2020-04-08
Canary

iruizmar:
Solving #125.

### DIFF
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -47,6 +47,16 @@ namespace UnityAtoms
         /// </summary>
         public E2 ChangedWithHistory;
 
+        /// <summary>
+        /// Whether Changed Event should be triggered on OnEnable or not
+        /// </summary>
+        public bool TriggerChangedOnOnEnable;
+
+        /// <summary>
+        /// Whether ChangedWithHistory Event should be triggered on OnEnable or not
+        /// </summary>
+        public bool TriggerChangedWithHistoryOnOnEnable;
+
         [SerializeField]
         private T _oldValue;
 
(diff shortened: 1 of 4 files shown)


Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -47,6 +47,16 @@ public abstract class AtomVariable<T, P, E1, E2, F> : AtomBaseVariable<T>, IGetE
...
 
+        /// <summary>
+        /// Whether Changed Event should be triggered on OnEnable or not
+        /// </summary>
+        public bool TriggerChangedOnOnEnable;
+
+        /// <summary>
+        /// Whether ChangedWithHistory Event should be triggered on OnEnable or not
+        /// </summary>
+        public bool TriggerChangedWithHistoryOnOnEnable;
```
AdamRamberg, line 58: Make the variable private, rename it to `_triggerChangedOnOnEnable` and add the attribute `[SerializeField]`.

Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -47,6 +47,16 @@ public abstract class AtomVariable<T, P, E1, E2, F> : AtomBaseVariable<T>, IGetE
         /// </summary>
         public E2 ChangedWithHistory;
 
+        /// <summary>
+        /// Whether Changed Event should be triggered on OnEnable or not
+        /// </summary>
+        public bool TriggerChangedOnOnEnable;
```
AdamRamberg, line 53: Make the variable private, rename it to `_triggerChangedOnOnEnable ` and add the attribute `[SerializeField]`.

Packages/Core/Runtime/Variables/AtomVariable.cs
AdamRamberg: Forgot initially to add that for clarity sake you should also explicitly set the value to `default` or `false`.

Packages/Core/Runtime/Variables/AtomVariable.cs
AdamRamberg: Forgot initially to add that for clarity sake you should also explicitly set the value to `default` or `false`.

iruizmar:
I forgot to leave a comment. I made the requested changes :). Let me know if there's something else.

AdamRamberg:
Looks good now! Merging the change to canary.

Issue #137, opened 2020-04-30
[BUG] Throwing NullReferenceException when updating TouchUserInput and no camera on scene


IceTrooper:
**Describe the bug**
When there is no camera on the scene and we're using somewhere UpdateTouchUserInput atom action to update our TouchUserInputVariable it throws tons of errors "NullReferenceException" because of this:
```
public Vector2 InputWorldPos { get { return Camera.main.ScreenToWorldPoint(InputPos); } }
```
and that:
```
public bool Equals(TouchUserInput other)
{
    return this.InputState == other.InputState && this.InputWorldPos == other.InputWorldPos && this.InputWorldPosLastFrame == other.InputWorldPosLastFrame && this.InputPosLastDown == other.InputPosLastDown;
}
```

Anyway why in Equals method we compare WorldPos and WorldPosLastFrame instead of Pos and PosLastFrame? That's the problem when we want to get InputWorldPos and we got no camera. It's ok to throw an error if we manually in script run InputWorldPos and no camera attached (first we should check if camera is on the scene), but it's bad when the internal method throws an error and we cannot do anything.

**To Reproduce**
Remove any camera on scene and this configuration on gameobject is enough to reproduce an error:
![obraz](https://user-images.githubusercontent.com/51023959/80756667-3e1e7d80-8b33-11ea-9025-af945a3c691f.png)

![obraz](https://user-images.githubusercontent.com/51023959/80756773-6c03c200-8b33-11ea-9738-dd2a95819a68.png)

If we want a simple fix, I'll do PR.

PR #139, opened 2020-05-01
#137 fixed throwing errors using TouchUserInputVariable...

IceTrooper:
... when no camera on scene

### DIFF
--- a/Packages/Mobile/Runtime/Types/TouchUserInput.cs
+++ b/Packages/Mobile/Runtime/Types/TouchUserInput.cs
@@ -81,7 +81,7 @@ namespace UnityAtoms.Mobile
         /// <returns>`true` if equal, otherwise `false`.</returns>
         public bool Equals(TouchUserInput other)
         {
-            return this.InputState == other.InputState && this.InputWorldPos == other.InputWorldPos && this.InputWorldPosLastFrame == other.InputWorldPosLastFrame && this.InputPosLastDown == other.InputPosLastDown;
+            return this.InputState == other.InputState && this.InputPos == other.InputPos && this.InputPosLastFrame == other.InputPosLastFrame && this.InputPosLastDown == other.InputPosLastDown;
         }
 
         /// <summary>
--- a/Packages/Mobile/Runtime/Types/TouchUserInput.cs
+++ b/Packages/Mobile/Runtime/Types/TouchUserInput.cs
@@ -24,54 +24,73 @@ namespace UnityAtoms.Mobile
         /// <summary>
         /// Current input state.
         /// </summary>
-        public State InputState;
+        public State TouchState { get => touchState; set => touchState = value; }
+        [SerializeField] private State touchState;
 
         /// <summary>
         /// Current input position.
         /// </summary>
-        public Vector2 InputPos;
+        public Vector2 Position { get => position; set => position = value; }
+        [SerializeField] private Vector2 position;
 
         /// <summary>
         /// Input position last frame.
         /// </summary>
-        public Vector2 InputPosLastFrame;
+        public Vector2 PositionLastFrame { get => positionLastFrame; set => positionLastFrame = value; }
+        [SerializeField] private Vector2 positionLastFrame;
 
         /// <summary>
         /// Input position last time the user pressed down.
         /// </summary>
-        public Vector2 InputPosLastDown;
-
-        /// <summary>
-        /// The input position in world space.
-        /// </summary>
-        /// <returns>The input position in world space.</returns>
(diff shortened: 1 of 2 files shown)


IceTrooper:
#137

soraphis:
Hm. I don't know enough about the Touch implementation.

The fix seems to be okay, but I think that: 

`public Vector2 InputWorldPos { get { return Camera.main.ScreenToWorldPoint(InputPos); } }`

needs to be changed. Camera.main is an expensive method:

> This property uses FindGameObjectsWithTag internally and doesn't cache the result. It is advised to cache the return value of Camera.main if it is used multiple times per frame.
https://docs.unity3d.com/ScriptReference/Camera-main.html

also, I think if `Camera.main == null` all `WorldPos` should return some kind of zero (e.g. Vector2.zero) Instead of throwing an exception. (but this is up to discussion)

IceTrooper:
Yes, you're right. I also think that something should change in InputWorldPosition property. It should return Vector2.zero if there isn't a camera on the scene. Based on @soraphis review I did some changes.

soraphis:
API has been changed. those are breaking changes and should be put in the upcoming changes for `UnityAtoms v5`

Packages/Mobile/Runtime/Types/TouchUserInput.cs
```diff
@@ -24,54 +24,73 @@ public enum State
         /// <summary>
         /// Current input state.
         /// </summary>
-        public State InputState;
+        public State TouchState { get => touchState; set => touchState = value; }
+        [SerializeField] private State touchState;
```
AdamRamberg, line 28: Use underscore as prefix. Eg. `_touchState`

Packages/Mobile/Runtime/Types/TouchUserInput.cs
```diff
@@ -24,54 +24,73 @@ public enum State
...
 
         /// <summary>
         /// Current input position.
         /// </summary>
-        public Vector2 InputPos;
+        public Vector2 Position { get => position; set => position = value; }
+        [SerializeField] private Vector2 position;
```
AdamRamberg, line 34: Use underscore as prefix. Eg. `_position`

Packages/Mobile/Runtime/Types/TouchUserInput.cs
```diff
@@ -24,54 +24,73 @@ public enum State
...
 
         /// <summary>
         /// Input position last frame.
         /// </summary>
-        public Vector2 InputPosLastFrame;
+        public Vector2 PositionLastFrame { get => positionLastFrame; set => positionLastFrame = value; }
+        [SerializeField] private Vector2 positionLastFrame;
```
AdamRamberg, line 40: Use underscore as prefix. Eg. `_positionLastFrame`

Packages/Mobile/Runtime/Types/TouchUserInput.cs
```diff
@@ -24,54 +24,73 @@ public enum State
...
 
         /// <summary>
         /// Input position last time the user pressed down.
         /// </summary>
-        public Vector2 InputPosLastDown;
+        public Vector2 PositionLastDown { get => positionLastDown; set => positionLastDown = value; }
+        [SerializeField] private Vector2 positionLastDown;
```
AdamRamberg, line 46: Use underscore as prefix. Eg. `_positionLastDown`

AdamRamberg:
Thanks for the PR! Like @soraphis said, this is a breaking change and we will need to wait adding this to Atoms until v5. FYI, the release of v5 of Atoms might take some time. The change for fixing the `Camera.main` issue could have been done in a none breaking change. So if you need this fix soon I would suggest 2 PRs, when for fixing `Camera.main` and one for do the refactoring of the prop / variable names etc. If no hurry, let the PR remain and we will add it to v5. 

Other than that I added some minior comments regarding this PR that needs to be fixed. Also created a new branch for v5 stuff that this PR will need to be merged to.

AdamRamberg:
@IceTrooper How is progress going on implementing the style fixes? When done I can merge this into the v5 branch.

IceTrooper:
Hey, there was a time I was busy, now I can make some changes. Tomorrow I will check it :)

soraphis:
since serialized fields were renamed, shouldn't we add [FormerlySerializedAsAttribute](https://docs.unity3d.com/ScriptReference/Serialization.FormerlySerializedAsAttribute.html) to them?

AdamRamberg:
@soraphis Yes, or only include this PR in v5, so merge it to the v5 branch instead of the canary branch.

Issue #136, opened 2020-04-29
[FEATURE] Typeless event for AtomBaseVariable


xvnnnk:
**Is your feature request related to a problem? Please describe.**
We can have AtomBaseVariable as a serialized field, but there is no way to access its `Changed` event. 
Since all AtomBaseVariables have Values and that value can change, it is possible to have a typeless Changed event.

**Describe the solution you'd like**
To have `TypelessChanged` event in `AtomBaseVariable`.

**Describe alternatives you've considered**
The only alternative is to manually track changes of AtomBaseVariable value every frame in script that uses it.

soraphis:
bind

soraphis:
> The only alternative is to manually track changes of AtomBaseVariable value every frame in script that uses it.

that is not true. not at all. see this example:

```
public class HealthBar : MonoBehaviour
{
    [SerializeField]
    private IntVariable MaxHealth;

    void Start()
    {
        MaxHealth.Changed.Register((oldValue) => this.DoSomething());
    }
    
    // todo: handle unregister!
   
    private void DoSomething()
    {
    }
}
```

a drawback of this method is, you need to hook it up via code. also you need to find a way of handling the unregister, better to save the lambda method in a variable. 

or just use a method with a parameter, but ignore the parameter.

you **do not** need to check every frame. 


other alternative: register a void event to raise when your TYPED-Event is raised

soraphis:
just checked, and the Register call allows `Action` as delegate.

so calling register via code you can use any `void method()` you want. 

in the example above this would mean:

        MaxHealth.Changed.Register(DoSomething);

mnarimani:
@soraphis 
This is the correct example:
```
public class Script : MonoBehaviour 
{
    public AtomBaseVariable Value;

    private void Start()
    {
        // How to register to AtomBaseVariable?
    }
}
```

soraphis:
> We can have AtomBaseVariable as a serialized field, but there is no way to access its Changed event.

AtomBaseVariable **has no** changed event. so there obviously is no way to access it.

mnarimani:
> > We can have AtomBaseVariable as a serialized field, but there is no way to access its Changed event.
> 
> AtomBaseVariable **has no** changed event. so there obviously is no way to access it.

Hence this feature request.
[...]

PR #140, opened 2020-05-01
Changes in GetEvent implementation to allow inheritance

soraphis:
brought up by this post here: https://github.com/AdamRamberg/unity-atoms/issues/136#issuecomment-622447714


AdamRamberg:
Looks great! Thanks!

Issue #133, opened 2020-04-28
[BUG] Code generation - Unchecking options has no effect on generated results


Robert-Oost:
**Describe the bug**
When using the code generator, all options are generated despite some of them being manually unchecked. 

**To Reproduce**

1. Open the Code generator
2. Uncheck some of the values. 
3. Click generate and wait a while.
4. Go to the folder with the generated files.

**Expected behavior**
Types that were unchecked are still generated.

**Screenshots**
This was my configuration:
![image](https://user-images.githubusercontent.com/39600335/80456262-95530100-892d-11ea-88a3-a2e4911fa4e7.png)
These were my results:
![image](https://user-images.githubusercontent.com/39600335/80456393-c92e2680-892d-11ea-97f1-12d30e8e9e6d.png)

**Environment**
 - OS: Windows 10
 - Unity Version 2019.3.11.f
 - Version of Unity Atoms [e.g. v4.1.0]

AdamRamberg:
Thanks for taking the time to report this issue! Will take a look at it when I have the time or if you (or someone else) is up for it then a PR would very much appreciated.

AdamRamberg:
Fixed in #147. Will be included in the next release of Atoms.

PR #143, opened 2020-05-03
Using OdinInspector to draw event's Raise button (#142)

### LINKED ISSUE #142: [Bug] Raise button doesn't show up in event inspector if class has no serializable fields.
**Describe the bug**
If the class that is used as event parameter doesn't have any serializable fields, then "Raise" button doesn't show up in the inspector.

**To Reproduce**
```
    [CreateAssetMenu]
    public class Example : AtomEvent<NoFieldClass>
    {
    }

    [Serializable]
    public class NoFieldClass
    {
        public string MyProperty { get; set; }
        public int OtherValue { get; set; }
    }
```
Create an instance of `Example` event.

**Expected behavior**
To have a Raise button that passes either an empty instance of class as event parameter (I prefer this) or passes null value. Whichever is better. The main point is having the "Raise" button.

**Screenshots**
image

mnarimani:
When creating new AtomEvent types, you also have to create new Editor script to draw "Raise" button for that new AtomEvent. 
With this PR, If you have OdinInspector imported in your project, Odin takes care of drawing "Raise" button for you for the types that don't have Editor script. 
If you don't have Odin, nothing will happen.
If you have custom Editor scripts, Odin won't touch them. So existing editors still work.

This fixes #142.

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -159,5 +159,13 @@ namespace UnityAtoms
                 }
             }
         }
+
+#if UNITY_EDITOR && ODIN_INSPECTOR
+        [Sirenix.OdinInspector.Button(Name = "Raise")]
+        private void OdinRaiseButton()
+        {
+            RaiseEditor(InspectorRaiseValue);
+        }
+#endif
     }
 }


AdamRamberg:
If you generate an AtomEvent using the [generator](https://adamramberg.github.io/unity-atoms/introduction/generator) then this is not needed. Or maybe I'm missing something?

mnarimani:
> If you generate an AtomEvent using the [generator](https://adamramberg.github.io/unity-atoms/introduction/generator) then this is not needed. Or maybe I'm missing something?

Oh, It's like I have missed something. I avoided using code generator because of #133 so I didn't know it would have fixed it. 

With that I don't know if it's still a good idea to have this feature for everybody.

For me though, I prefer to have it for my own because it lets me to have one less script in the project for every custom event type.

Decision is yours. I'm ok with merge or close.

AdamRamberg:
Cool! I will close this one then since it is already covered by Atoms.

Issue #141, opened 2020-05-02
[BUG] Duplication check in ValueLists


mnarimani:
**Describe the bug**
ValueLists implement `IList<T>` not `ISet<T>`. So they should behave like `List<T>` not `HashSet<T>`. They should have no problem with having duplicate elements in them. 
This line:
https://github.com/AdamRamberg/unity-atoms/blob/ce9e818ac3d9315364ebf539926597a77a5adb85/Packages/Core/Runtime/ValueLists/AtomValueList.cs#L51
Should be removed.

AdamRamberg:
Hi @mnarimani! Thanks for this issue. This change sounds sane. Would you mind submitting a PR for this against the canary branch?

mnarimani:
> Hi @mnarimani! Thanks for this issue. This change sounds sane. Would you mind submitting a PR for this against the canary branch?

Sure.

AdamRamberg:
Fixed in #144. Closing the issue.

PR #144, opened 2020-05-03
(#141) Removed duplicate check before adding an item to the list

mnarimani:


mnarimani:
Oh, I didn't notice this includes my last change. I will create new PR.

PR #145, opened 2020-05-03
(#141) Removed duplicate check before adding an item to the list

mnarimani:


AdamRamberg:
Looks good! Thanks! Will be included in next release of Unity Atoms.

PR #152, opened 2020-05-08
Add the [Serializable] attribute in the doc

bguyl:
Add the [Serializable] attribute in the documentation for the generator example


AdamRamberg:
Thanks for the PR! Looks good and will merge. The website will be updated when we release the next version of Unity Atoms.

Issue #90, opened 2019-11-15
[FEATURE] Way to find usages on Events. Who's calling this and Who's listening to this


RErwandi:
I'm previously using UniRx for event subscription and Atoms seems a great alternative. The way events saved as ScriptableObject decoupled a lot of things and allow other programmers to see game events available on the project. But one crucial thing is missing when i'm using this way. I lost/don't know how to track game events. Where is this called and Who is listening to this. In scripts, we can use Find Usages(F12) and other scripts that refer to that event is shown.

The solution might be to serialize/show in inspector of game event the list of its current action so we can see what's currently listening to that game event.
Also show debug log when event is raised so we can use console stack trace to see who's raised the event.

AdamRamberg:
I think this is a great feature to add! Would be happy to see a PR for this.

puschie286:
hi, we are currently developing such a event tracker - its not done yet but in a prototype stage
https://github.com/puschie286/unity-atoms/tree/eventtracker

it shows a list of all fired events (in this session)
buttons for selecting the specific event, listeners
calling info: value + callstack

AdamRamberg:
@puschie286 Any news on the event tracker?

puschie286:
a tiny one - we had a few problems with the UNITY_EDITOR directive when building ios and android builds, so we developed a version with an different approach - feel free to use anything of it.
https://github.com/puschie286/unity-atoms/tree/feature/eventtracker
currently a few basic features ( like clear button, filter option to exclude events )

miikalo:
Hello @puschie286 , some interest on Discord channel is starting to develop for this kind of feature, do you happen to have any updates?

puschie286:
hi @miikalo 
we integrated a event catalog that list all event objects and event types ( with a few additional information ) - this does not include variable usage currently.
you can see it in the newest commit on https://github.com/puschie286/unity-atoms/tree/feature/eventtracker
the event log got auto scroll and a clear button, so still very basic but great to understand the flow of the game and reconstruct action ( or finding missing events )
current known issues are: the last event is not shown properly and auto scroll use current scroll position

RuchirRaj:
Any update on this? Still holding back on integrating this to my project because of this

puschie286:
Unfortunately not - we still use it in the some Version, would be nice if you could integrate some parts :)

soraphis:
Closing v5 PRs as there is no active development on v5 anymore

PR #159, opened 2020-05-15
Debug events by displaying stack traces for events

AdamRamberg:
This PR adds the feature of debugging events by displaying stack traces for the handlers it calls. It implements #90. The feature is heavily inspired and based on how [Sciptable Object Architecture](https://github.com/DanielEverland/ScriptableObject-Architecture) does the same thing, so credits to [Daniel Everland](https://github.com/DanielEverland) for that. 

The stack event looks like this in the editor for events: 
Skärmavbild 2020-05-15 kl  12 33 59

This feature is by default turned off (because of the performance penalty of using it). The feature can be turned on via Unity's preferecnes: 
preferences-overview

Another thing to note is that all stack traces are stored in a static variable in a static class. All calls for adding stack traces are toggled with the compilation flag `UNITY_EDITOR`. The reason for this is to eliminate any penalty at runtime of your build for using this feature.


soraphis:
> Another thing to note is that all stack traces are stored in a static variable in a static class. All calls for adding stack traces are toggled with the compilation flag UNITY_EDITOR. The reason for this is to eliminate any penalty at runtime of your build for using this feature.

might want to use https://docs.microsoft.com/de-de/dotnet/api/system.diagnostics.conditionalattribute?view=netcore-3.1

AdamRamberg:
> might want to use https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=netcore-3.1

Didn't know about this one. Think what we do is basically the same, but nice to know about conditional attribute for the future. 

> for random id's you could generate a UUID instead of a random int

The PR is using the `GetInstanceID()` of the Event (Scriptable Object) as key for the stack traces. This should be sufficient in my mind.

soraphis:
I havn't yet found time for a proper review, just glimpsed over the code, what i meant with random is here: 

https://github.com/AdamRamberg/unity-atoms/pull/159/files#diff-887e1b37c3ef84c706d938bed14496c5R16

AdamRamberg:
> I havn't yet found time for a proper review, just glimpsed over the code, what i meant with random is here:
> 
> https://github.com/AdamRamberg/unity-atoms/pull/159/files#diff-887e1b37c3ef84c706d938bed14496c5R16

Aha, that makes sense! Just pushed a new commit using GUID instead of a random int. I also changed the Label in the detailed stack trace view to a read only TextField, which makes it possible to copy and paste the stack trace.

Packages/Core/Runtime/StackTrace/StackTraceEntry.cs
```diff
@@ -0,0 +1,66 @@
...
+                _frameCount = Time.frameCount;
+            }
+        }
+
+        public static StackTraceEntry Create(object obj) => new StackTraceEntry(Environment.StackTrace, obj);
+
+        public static StackTraceEntry Create() => new StackTraceEntry(Environment.StackTrace);
+
+
+        public override bool Equals(object obj)
```
soraphis, line 42: would write as:

```Csharp

public override bool Equals(object obj) => Equals(obj as StackTraceEntry);

public bool Equals(StackTraceEntry other) => other != null && other._id == _id;
// or even: 
public bool Equals(StackTraceEntry other) => other?._id == _id;
```

Packages/Core/Runtime/StackTrace/StackTraceEntry.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public override int GetHashCode() => _id.GetHashCode();
+
+
+        public override string ToString() =>
+            _constructedWithValue ?
+                $"{_frameCount}   [{(_value == null ? "null" : _value.ToString())}] {_stackTrace}" :
+                $"{_frameCount} {_stackTrace}";
+
+        public static implicit operator string(StackTraceEntry trace) => trace.ToString();
```
soraphis, line 64: implicit operators should be used with caution ... I don't know how to feel about this one.

(causes heap allocations (string concatenation), depends on the ToString implementation of _value, which could be anything) 
https://softwareengineering.stackexchange.com/a/284377

Packages/Core/Runtime/StackTrace/StackTraceEntry.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            if (Application.isPlaying)
+            {
+                _frameCount = Time.frameCount;
+            }
+        }
+
+        public static StackTraceEntry Create(object obj) => new StackTraceEntry(Environment.StackTrace, obj);
+
+        public static StackTraceEntry Create() => new StackTraceEntry(Environment.StackTrace);
```
soraphis, line 39: probably better use: `new StacktTrace(4).ToString()` (will skip 4 frames, which are always: 1. get_StackTrace() / or cctor, 2. this line itself, 3. EventBase:Raise(), 4. Event.Raise(T item)`

this might be a different frame count for both Create methods, but it will Atom internal stuff, which is always the same

Packages/Core/Editor/Extensions/StringExtensions.cs
```diff
@@ -43,5 +43,7 @@ public static string Capitalize(this string str)
             a[0] = char.ToUpper(a[0]);
             return new string(a);
         }
+
+        public static string GetFirstLine(this string str) => str.Split(new[] { '\r', '\n' }).FirstOrDefault();
```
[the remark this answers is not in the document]
AdamRamberg: Fixed in last push.

Packages/Core/Runtime/Preferences/AtomsPreferences.cs
```diff
@@ -0,0 +1,99 @@
+﻿#if UNITY_EDITOR
```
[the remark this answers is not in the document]
AdamRamberg: No. Since AtomPreferences contains properties that needs to be accessed by classes in the runtime assembly (eg. `IsDebugModeEnabled`) it also needs to reside there.

Packages/Core/Runtime/StackTrace/StackTraceEntry.cs
AdamRamberg: Good one! Fixed in last commit.

Packages/Core/Runtime/StackTrace/StackTraceEntry.cs
AdamRamberg: Yea, this is probably just unnecessary. Removed the implicit conversion operator and I'm instead using `ToString` explicitly when needed.

Packages/Core/Runtime/StackTrace/StackTraceEntry.cs
AdamRamberg: Fixed. Also added skipFrames as a parameter to the `Create` method. Since the StackTrace is created here in this method the correct amount of stack traces to shave off was 3 instead of 4.

AdamRamberg:
@soraphis What do you say? Good to merge? 

FYI. When this PR is OK:ed I'm planning to release v4.2. Let me know if there is something else we need to get in before making the release.

AdamRamberg:
Since there've been activity in quite some time on this one I will merge it to canary.

PR #167, opened 2020-06-23
Make patch check OS agnostic.

jmacgill:


jmacgill:
Will reopen with just the changes for this fix.

PR #169, opened 2020-06-25
Generator creates compilable code with type specific modifications

jmacgill:
Adds support for nested IF blocks
Adds TYPE_IS_X condition
Adds IS_NUMERIC IS_VECTOR IS_COLLIDER
Updates templates using the above so that generated code matches repository code

Some whitespace diffs remain.

### DIFF
--- a/Packages/BaseAtoms/Runtime/Events/VoidEvent.cs
+++ b/Packages/BaseAtoms/Runtime/Events/VoidEvent.cs
@@ -8,10 +8,9 @@ namespace UnityAtoms.BaseAtoms
     [EditorIcon("atom-icon-cherry")]
     [CreateAssetMenu(menuName = "Unity Atoms/Events/Void", fileName = "VoidEvent")]
     public sealed class VoidEvent : AtomEvent<Void> {
-                public override void Raise()
+        public override void Raise()
         {
             Raise(new Void());
         }
-        
      }
 }
--- a/Packages/BaseAtoms/Runtime/Events/VoidEvent.cs
+++ b/Packages/BaseAtoms/Runtime/Events/VoidEvent.cs
@@ -7,10 +7,10 @@ namespace UnityAtoms.BaseAtoms
     /// </summary>
     [EditorIcon("atom-icon-cherry")]
     [CreateAssetMenu(menuName = "Unity Atoms/Events/Void", fileName = "VoidEvent")]
-    public sealed class VoidEvent : AtomEvent<Void> {
-                public override void Raise()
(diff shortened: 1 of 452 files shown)


AdamRamberg:
Thanks for the PR! 🙏 

I will take a look at it when I get the time. Hopefully soon-ish, but no promises.

jrmacgill:
If it helps I can break it into smaller parts.

AdamRamberg:
@jrmacgill It's fine as it is. I just haven't had the time to go through it yet. Will try to make time this week.

Packages/BaseAtoms/Runtime/Events/VoidEvent.cs
```diff
@@ -7,8 +7,7 @@ namespace UnityAtoms.BaseAtoms
     /// </summary>
     [EditorIcon("atom-icon-cherry")]
     [CreateAssetMenu(menuName = "Unity Atoms/Events/Void", fileName = "VoidEvent")]
-    public sealed class VoidEvent : AtomEvent<Void>
-    {
+    public sealed class VoidEvent : AtomEvent<Void>     {
```
soraphis, line 10: is not valid based on our https://github.com/AdamRamberg/unity-atoms/blob/master/.editorconfig

make sure your IDE uses the editorconfig (vs code needs a plugin for that IIRC)

soraphis: invalid formatting makes up most of the changeset. please correct.

jrmacgill: Yeah - I need to fix this in the generator mostly - not much point in having it generate code that does not need to be modified if it then relies on the IDE to fix the formatting anyway.  I was having trouble getting linefeeds and whitespace to behave where blocks were removed.

Failing that, I could remove the generated files from the PR - I'll see if I can avoid that though.  I'll take a run at it this weekend.

AdamRamberg:
@jmacgill Looked through the PR and it's mostly all good. Made a couple of changes myself (hope you don't mind) to fix the issues that @soraphis were pointing out. 

Note, the logic to determine if a conditional is inline or not works for now, but could become problematic in the future for other use cases (see my comment in the code in `Templating.cs`). This is fine for me and we can cross this bridge if / when we get there. 

Will merge this PR. Thanks for contributing 🙏

jrmacgill:
Thanks for the cleanup changes! (Don't mind at all!)

Yeah the inline check is... fragile.  The comment is good, lets see if/how templates evolve to see what would be more robust.

PR #155, opened 2020-05-10
Canary - Fix #153 

bguyl:

PR #174, opened 2020-06-29
Update AtomBaseReferenceDrawer.cs

soraphis:
the changes introduced by #155 neglected that fields with visible children could use a custom drawer to render in a single line

image

this patch addresses the problem

image


soraphis:
if this request will be accepted, i'd suggest the "squash and merge" strategy

AdamRamberg:
Thanks for the PR @soraphis! I think this is a good enough work around for now. Noticed a couple of glitches when trying the drawer out, like for example the foldout arrow being hidden by the "usage" menu button for atoms with children (eg. `TouchUserInputReference`). This is however nothing major and only cosmetic and was not either introduced by this PR. Will merge this to canary.

Issue #176, opened 2020-07-06
[BUG] Sub FSM doesn't update value in primary FSM


TriangularCube:
**Describe the bug**
When using sub FSM in a state which is  _not_ the initial state of the primary FSM, the primary FSM's Value doesn't not update based on the sub FSM.

**To Reproduce**
1. Create two FSMs.
2. Create at least two states in the primary FSM, then attach the other FSM as a sub FSM in slot 2.
3. Create at least two states in the sub FSM.
4. In the primary FSM, switch to the state using the sub FSM.
5. In the sub FSM, switch state.

**Expected behavior**
The primary FSM's Value should update according to the sub FSM's state. This is apparently not an issue if the sub FSM is the primary FSM's initial state.

Here's a repo demonstrating the problem.
https://github.com/TriangularCube/fsm-demo

**Environment**
 - OS: Windows 10 2004
 - Unity Version: 2019.4.0f1
 - Version of Unity Atoms: 4.2.1

TriangularCube:
Some more information, if the sub FSM is attached to the primary FSM's initial value, it will work until the primary FSM switches states, then it will fail to update value from sub FSM when switching back.

AdamRamberg:
Hi there @TriangularCube!

Thanks for submitting this issue! I could take a look at it when I get the time (not 100% sure when though since my time is limited as of lately). If you want this fixed soon I would encourage you to dig into the issue and make a PR.

TriangularCube:
You got it!

AdamRamberg:
Fixed in #178. Closing the issue.

PR #178, opened 2020-07-12
Fix Sub FSM Value issue

TriangularCube:
Fix for #176.

Turns out when updating FSMs, if a state has a Sub FSM, its Value is updated to be the Sub FSM's Value, and not the base FSM's Value, thereby rendering `submachine` null.


AdamRamberg:
Thanks for the PR! Really nice to test when you provided a demo repo ⭐ Looks good to me. This fix will be included in the next release of Unity Atoms. If you want to use this before the release you can use the [Github url method](https://adamramberg.github.io/unity-atoms/introduction/quick-start#github-url) when including Atoms in your project and then reference the canary branch (using `#canary`).

PR #182, opened 2020-07-18
Add workaround for "Enter Play Mode" feature to enable initial values

hazarartuner:
This may not be the optimal solution but It seems working and we can think this as a workaround until we find a better approach.

**Important Note:** If you enable the "Enter Pay Mode" feature in Unity, you have to restart the editor to make this workaround work.

### DIFF
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -2,7 +2,9 @@ using System;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Serialization;
+#if UNITY_EDITOR
 using UnityEditor;
+#endif
 
 namespace UnityAtoms
 {
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Serialization;
+using UnityEditor;
 
 namespace UnityAtoms
 {
@@ -105,6 +106,13 @@ namespace UnityAtoms
         [SerializeField]
         private T _initialValue = default(T);
 
+#if UNITY_EDITOR
+        /// <summary>
+        /// List of all AtomVariable instances in editor.
+        /// </summary>
+        private static List<AtomVariable<T, P, E1, E2, F>> _instances = new List<AtomVariable<T, P, E1, E2, F>>();
+#endif
+
         /// <summary>
         /// When setting the value of a Variable the new value will be piped through all the pre change transformers, which allows you to create custom logic and restriction on for example what values can be set for this Variable.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Packages/Core/Runtime/Variables/AtomVariable.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEditor;
```
soraphis, line 4: needs to be wrapped in `#if UNITY_EDITOR` or the unity project will not build

hazarartuner:
> are lists the only things affected? wouldn't we need those changes on basically all atom types?

Actually it should affect all atom variables, not just atom lists. It collects all atom instances in a List maybe this looks to you like it is working only for lists but it is not.

soraphis:
> > are lists the only things affected? wouldn't we need those changes on basically all atom types?
> 
> Actually it should affect all atom variables, not just atom lists. It collects all atom instances in a List maybe this looks to you like it is working only for lists but it is not.

oh, dang. you're right. I let the `_instances` variable confuse me. sorry, my bad

hazarartuner:
> > > are lists the only things affected? wouldn't we need those changes on basically all atom types?
> > 
> > 
> > Actually it should affect all atom variables, not just atom lists. It collects all atom instances in a List maybe this looks to you like it is working only for lists but it is not.
> 
> oh, dang. you're right. I let the `_instances` variable confuse me. sorry, my bad.
> 
> Why not using a HashSet instead of a list? you wouldn't need to remove+add

You are right :) I've replaced the List type with HashSet :D

Packages/Core/Runtime/Variables/AtomVariable.cs
AdamRamberg: Have been addressed in the second commit

AdamRamberg:
@hazarartuner Thanks for this PR! The PR looks good. I would like however you to do 2 things before I will merge this change: 
1) Could you rebase with current canary changes. 
2) Would like you to add this feature under a feature flag for now, since it's considered a work around and will very likely be changed in a future minor release of Unity Atoms. I want you to add a flag similar to `IsDebugModeEnabled` in `AtomPreferences` and call it something like "Enter Play Mode (Experimental)" / `IsEnterPlayModeEnabled`. Then use that variable to toggle the code that you added in your previous commits. 

Let me know if you got any questions!

soraphis:
> Would like you to add this feature under a feature flag for now...

not sure if this is necessary, since it already uses `if (EditorSettings.enterPlayModeOptionsEnabled)`. so users already need to toggle one experimental flag in the settings, which will break unity atoms, leaving the feature broken until a second flag is toggled on seems to be quite redundant.

@hazarartuner could you please rebase this feature onto the current canary? thanks.

hazarartuner:
@soraphis @AdamRamberg I rebased it. Is it enough?

soraphis:
@AdamRamberg @miikalo I'd accept this PR even though there is not a Atom Specific "feature toggle" in the settings, because I think toggling the feature itself in unity should be enough. 

what do you think (if I receive no answer till tomorrow, I'll accept the PR. the bene)

miikalo:
> @AdamRamberg @miikalo I'd accept this PR even though there is not a Atom Specific "feature toggle" in the settings, because I think toggling the feature itself in unity should be enough.
> 
> what do you think (if I receive no answer till tomorrow, I'll accept the PR)

Without workaround: "Enter play mode" enabled in Unity, Atoms breaks, feature disabled, Atoms works.
With workaround: "Enter play mode" enabled in Unity, Atoms works, feature disabled, still works.

So if the workaround does not bother Unity when the "Enter play mode" feature is disabled, everything should be fine. I agree with @soraphis that an additional flag would confuse things here.

TowerCrow:
Does this PR also addresses Start Cleared option on ValueLists? Because that also doesn't work when the new Enter Play Mode settings are enabled.

PR #184, opened 2020-08-01
Creating new atom in current directory improvement

IceTrooper:
It's annoying when new Atom, Event is created in root Asset directory. It's a small improvement that adds support for creating a new .asset file in the current ScriptableObject directory if SO is selected. If GameObject on the scene is selected it creates in root Asset folder, because it's impossible to detect directory.
#185


AdamRamberg:
Thanks for this PR! Looks good and will merge :)

Issue #189, opened 2020-08-13
[FEATURE] Tutorials improvements


miikalo:
**Is your feature request related to a problem? Please describe.**
Tutorials are important in driving new user adaptation. The current state of tutorials could use some reworking in structure and content.

**Describe the solution you'd like**
After having a clear starting point with the existing tutorials, I felt abandoned by the tutorials soon after the very basics and was left to my own devices to figure out the features. This can be a huge turn-off to new users and big wins can be made with relatively little effort. Tutorials save the new users time and reduce the risk of misunderstanding a feature and misusing it, then dismissing the feature as unimportant or unhelpful or working around intended, misunderstood functionality somehow.

**Describe alternatives you've considered**
Leave the tutorials on the website be and create a YouTube series of in-depth tutorials.

**Additional context**
Several people in the #help channel in Discord have been asking for variable instancer tutorials.

AdamRamberg:
Thanks for submitting this PR! Really nice work 💪 After my first read through of the changes my biggest concern was where the Mono Hooks section is located combined with the content of that section. It’s assuming that the basic tutorials are right before it. I would suggest moving it to the basic tutorial section or rewriting it so that it’s building on the other intermediate sections. The easiest would probably be to just move it to the basic section. 

Will read through the docs more thoroughly again next chance I get for more detailed feedback.

PR #190, opened 2020-08-13
#189 Restructure tutorials and add two new tutorials

miikalo:
Addresses #189 

- Non-generated documentation review and edit
  - Consistent terminology
  - Stylistic changes
  - Minor structure changes
  - Minor clarifications
  - Typo fixes

- Split basic tutorial into two
- Tutorials are easier to follow when they are short and to the point
- Added event and variable instancer tutorials
- Had to bump node version for the docker container to work

Please let me know if the edits are too opinionated.


AdamRamberg:
Thanks for submitting this PR! Really nice work 💪 After my first read through of the changes my biggest concern was where the Mono Hooks section is located combined with the content of that section. It’s assuming that the basic tutorials are right before it. I would suggest moving it to the basic tutorial section or rewriting it so that it’s building on the other intermediate sections. The easiest would probably be to just move it to the basic section.

Will read through the docs more thoroughly again next chance I get for more detailed feedback.

miikalo:
Pushed an amend where Mono Hooks moves to Basics and the last paragraph from Actions moves to Mono Hooks.

I have a question about this part in the "Overview":

Unity Atoms is an event based system that encourages the game to be as data-driven as possible. The four most fundamental parts of Unity Atoms are:
- Data
- Events
- Listeners
- Functions
- Collections

Which four of the five should stay on the list, or just fix the "four" to "five"? I would also remove a mention of the "four fundamentals" from the end of the Mono Hooks tutorial later because the tutorials are not named after the fundamentals and "Data" for example would need two separate tutorials (Variables, Collections).

Issue #52, opened 2019-08-12
Add Where condition to Listeners, eg. call handlers every Nth time


IceTrooper:
I got a new small idea to add, I can make a pull request when two earlier issues will be resolved.
What do you think about adding two properties to GameEventListener and one mechanic:
1. int every (default: 1)
2. bool triggerFirst (default: false)
3. some private int counter

We could trigger some UnityEvents/GameActions every few raised Event, not every once (like it's now). "triggerFirst" means if we set "every=5", will something be done on the first occurrence or not). Every raised event a counter will be increased to check how many times it triggered. It would add more flexibility to Listeners and it's a small change.

soraphis:
[something] ----raises ----> [Atoms Event] ---- notifies ----> [listener]

the listener itself is a developer-class not a library-class. a really small subset of listeners will use this logic. for the cases where they are needed a developer could just add 3 lines of code:

    [SerializeField][HideInInspector] private int counter;
    [SerializeField] repeatAmount;

    .... some code, and first line of the listener method:
 
        if( ( (++counter) % repeatAmount ) != 0 ) return;


I - personally - don't see the need to include it. But I might be wrong, and the demand for this functionality is higher than i thought

IceTrooper:
What do you mean it's developer-class not library-class?

@soraphis Where do you add those 3 lines of codes? I wanted to use an Int Listener/Float Listener etc with that functionality, so I can't modify their code, because it's compiled into a package.

AdamRamberg:
I think that instead of adding this specific logic to the Listener it would be better to add a BoolFunction to the Listener that you can apply your own logic to, eg.`responseCondition`. Wouldn't mind a PR for this.

AdamRamberg:
**Disregard my previous comment.** 

**What I think we should do**
Instead of doing a specific solution (call handlers only every nth time) we should do a general solution. I'm thinking that we should add a "Where" condition to Listeners. It would be a `AtomFunction<bool>`, `AtomFunction<bool, T1>` or `AtomFunction<bool, T1, T2>` depending on if the condition should care about the data being sent in or not. The AtomFunction will basically just implement its own logic for if the Event should be let through and the handlers / responses should be called.

We could also add some basic conditions, for example: 
- Every nth time (`AtomFunction<bool>`)
- If a Variable (must not be "related" to the Listener) has a value (`AtomFunction<bool>`)

[...]

PR #191, opened 2020-08-31
#52 Basic implementation of conditions for listeners

miikalo:
Addresses #52 to spark more discussion about the topic.

This brings in condition types that can be extended with custom logic.

Create a script extending the Condition type of your choice, here `Int`:
image

Create instances of that Condition:
image

Use in a listener of that type:
image

Missing:
- Icon for Conditions
- Condition types for default types that Atoms ships with

Improvements:
- Ideally the Conditions in the list of the inspector of the listener would have Condition references, where you can choose between a serialized instance editable in the inspector, or an asset, like it is now
- More elegant way of handling the operator selection
- Regenerate docs


Packages/Core/Runtime/Conditions/AtomCondition.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace UnityAtoms
+{
+    /// <summary>
+    /// Base abstract class for Conditions. Inherits from `AtomFunction` and 
+    /// must implement `Condition` Predicate
+    /// </summary>
+    public abstract class AtomCondition<T> : AtomFunction<T, T>
+    {
+        public Predicate<T> Predicate;
```
soraphis, line 12: Should actually be `AtomCondition<T> : AtomFunction<bool, T>`
and since AtomFunction already defines a `public Func` this is already in existence here, so the Predicate<T> delegate is not needed.

Packages/Core/Runtime/Conditions/AtomCondition.cs
```diff
@@ -0,0 +1,17 @@
...
+
+        public AtomCondition() => Predicate = Condition;
+        public abstract bool Condition(T value);
```
soraphis, line 15: what is the purpose of Condition? is it used? have not seen the usage of it yet.

soraphis (commented):
for the condition list:
conditions could be coded in DNF to do so, a struct would be needed e.g. 

```
public struct ConditionBlock<T>{
    List<AtomCondition<T>> Conditions; // all conditions are AND combined.

    public bool All() => Conditions.All(x => x.Func());
}
```

Within the AtomBaseListener a `List<ConditionBlock<<T>> Conditions` would be used and all condition blocks are OR combined.
so the check in the code would be:

`Conditions.Any(x => x.All());` 

every boolean logic can be expressed in DNF this way the operator field would not be needed.

note: an option to toggle from DNF to KNF could be nice, but DNF usually feels pretty natural. 

_____

> Ideally the Conditions in the list of the inspector of the listener would have Condition references, where you can choose between a serialized instance editable in the inspector, or an asset, like it is now

can you explain what you mean by "a serialized instance"? isn't an asset (aka scriptable object) a "serialized instance"?
I think for AtomFunctions we only have assets, the concept of reference is IIRC exclusive for Variables.

a switch between assets and `UnityAction` would be possible, but this would be quite uncomfortable in editor-style-wise

Packages/Core/Runtime/Conditions/AtomCondition.cs
miikalo: See the first picture in the PR description, it is the actual condition evaluation method to implement which gets assigned to the Predicate. Each extended <Type>Condition needs their own implementation of Condition(T value) to work. I went this way because Predicate <T>is a delegate. If it looks, odd, it's because I've never used it, maybe I got it wrong.

soraphis: AtomFunction<T> defines a method `virtual T Call`, because AtomCondition<T> should inherit from AtomFunction<bool, T> the implementation should override the method `bool Call(T t1)`

you can see this in: https://github.com/AdamRamberg/unity-atoms/blob/master/Packages/BaseAtoms/Runtime/Functions/Clamp/ClampInt.cs

miikalo: Ah, now I get it, thanks!

miikalo:
No longer uses a Predicate, makes things much simpler. I added the Condition to the Generator, but I'm not sure how to regenerate the existing base atoms (int, bool, string, etc.), using "regenerate all" generates a ton of files and breaks everything, not sure what's going on.

Not at all sure what to do with the ConditionBlock to be honest, how would the user choose the logical operator if something like that is used?

Packages/Core/Runtime/Listeners/AtomBaseListener.cs
```diff
@@ -69,20 +83,36 @@ private void OnDisable()
...
+            bool shouldRespond = _operator == AtomConditionOperators.And ? true : false;
+
+            // Evaluate conditions and decide whether to respond or not
+            for (int i = 0; _conditions != null && i < _conditions.Count; ++i)
             {
-                var action = _actionResponses[i];
+                var condition = _conditions[i];
 
-                if (action == null) continue;
+                if(condition == null) continue; // TODO: Use IsUnassigned when available
```
soraphis, line 93: Blocked by #202

Packages/Core/Runtime/Listeners/AtomBaseListener.cs
```diff
@@ -69,20 +83,36 @@ private void OnDisable()
...
 
-                if (action is AtomAction<T> actionWithParam)
-                {
-                    actionWithParam.Do(item);
-                }
-                else
+                if(_operator == AtomConditionOperators.And) shouldRespond &= condition.Call(item);
+                if(_operator == AtomConditionOperators.Or) shouldRespond |= condition.Call(item);
```
soraphis, line 96: early return would be possible

    if(_operator == AtomConditionOperators.And && !shouldRespond) break;
    if(_operator == AtomConditionOperators.Or  &&  shouldRespond) break;

Packages/Core/Runtime/Listeners/AtomBaseListener.cs
```diff
@@ -69,20 +83,36 @@ private void OnDisable()
...
+            }
+            
+            if(shouldRespond)
+            {
```
soraphis, line 100: flip if direction to skip nesting, actually remove "shouldRespond" alltogether, using a negated logic and a `return` in the loop above

miikalo:
Should now include all .meta files as well for the generated conditions, also fixed accidentally removed call to the actual condition (!) in the loop, now works as intended again.

Tommie0607:
Hey @miikalo , first of all thanks for the conditions-logic implementation. 
Is it possible the List<AtomCondition<T>> in AtomBaseListener doesn't serialize (because its a generic type) and thus doesn't show up in the inspector... or is it just me?

miikalo:
Hi @Tommie0607 , I'm mostly using 2020.1, but I tested this on 2019.4 LTS and it works. There's an upcoming tutorial here: https://github.com/miikalo/unity-atoms/blob/feature/conditions-tutorial/docs/tutorials/conditions.md

The tutorial is associated with the PR #206 

Here's a screenshot of it working in 2019.4:

image

PR #193, opened 2020-09-15
Adding a little bit of clarity to installation

mutmedia:
Adding a little bit of clarity to installation for someone completely new to unity packages


miikalo:
I have nothing against this, but the PR should target canary instead.

PR #202, opened 2020-10-07
Added for AtomReference a check to see if an Atom is actually assigned.

soraphis:
Some scripts or checks might be used to see whether or not a Reference has been assigned.

this method allows to do so.


miikalo (approved):
Looks good to me.

PR #212, opened 2020-11-27
Unity Atoms Input System Integration

Casey-Hofland:
This is just a quick demo to suggest to the Unity Atoms team. There 'was' a line on the github that said to simply contact you... but creating a Pull Request is the more funner way of doing things.

Here we have a [YouTube link](https://www.youtube.com/watch?v=q7W8FyTIriQ&feature=youtu.be) that explains what these scripts are for. The scene is a bit bare bones since testbed and all that.


Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+    where T : struct
+    where P : struct, IPair<T>
+    where C : AtomBaseVariable<T>
+    where V : AtomVariable<T, P, E1, E2, F>
+    where E1 : AtomEvent<T>
+    where E2 : AtomEvent<P>
+    where F : AtomFunction<T, T>
+    where VI : AtomVariableInstancer<V, P, T, E1, E2, F>
+{
+    public E1 any;
```
AdamRamberg, line 14: Make variable private and use the SerializeField attribute. Change the name to have a leading undescore. Eg. `_any`.

Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+    public E1 started;
```
AdamRamberg, line 15: Same as above

Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+    public E1 performed;
```
AdamRamberg, line 16: Same as above

Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+    public E1 canceled;
```
AdamRamberg, line 17: Same as above

Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+    public E1 waiting;
```
AdamRamberg, line 18: Same as above

Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+    public E1 disabled;
```
AdamRamberg, line 19: Same as above

Packages/InputSystem/Runtime/CallbackContextEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    public V variable;
```
AdamRamberg, line 21: Same as above

Packages/InputSystem/Runtime/CallbackContextEventVector2.cs
```diff
@@ -0,0 +1,7 @@
+using UnityAtoms.BaseAtoms;
+using UnityEngine;
+
+[CreateAssetMenu(menuName = "Unity Atoms/Examples/CallbackContext Vector2")]
+public class CallbackContextEventVector2 : CallbackContextEvent<Vector2, Vector2Pair, Vector2Constant, Vector2Variable, Vector2Event, Vector2PairEvent, Vector2Vector2Function, Vector2VariableInstancer>
```
AdamRamberg, line 5: Rename it to `Vector2CallbackContextEvent` in order to follow the naming conventions of other atoms, eg. Variables.

Packages/InputSystem/Runtime/CallbackContextEventVector2.cs
```diff
@@ -0,0 +1,7 @@
+using UnityAtoms.BaseAtoms;
+using UnityEngine;
+
+[CreateAssetMenu(menuName = "Unity Atoms/Examples/CallbackContext Vector2")]
```
AdamRamberg, line 4: Change to: `Unity Atoms/Input System/Vector2 Callback Context`

Packages/InputSystem/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "com.caseydecoder.unity-atoms-input-system",
```
AdamRamberg, line 2: If it's supposed to be merged to the main project it should follow the same conventions as other packages, eg: `com.mambojambostudios.unity-atoms-input-system`.

Packages/InputSystem/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "displayName": "Unity Atoms Input System",
+  "version": "4.3.0",
+  "unity": "2018.3",
+  "description": "Bla Bloe Bladiebla",
```
AdamRamberg, line 6: Maybe the description should be a little bit more descriptive ;). Eg: Use Unity Atoms with Unity's Input System

Packages/InputSystem/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "keywords": [
+    "scriptable objects",
+    "unity"
+  ],
+  "category": "Unity",
+  "files": [
+    "/package.json",
+    "/package.json.meta",
+    "/Runtime",
+    "/Runtime.meta",
```
AdamRamberg, line 16: Remove trailing comma

AdamRamberg (changes requested):
Thanks a bunch for this PR! Simple but yet very useful addition to Atoms. Awesome that you submitted a YT-video as well :D 

I added some requests for changes to the PR. Most of it is just to align it with the style of the rest of the project. If you fix those I will merge it to the canary branch and it will be available in a future release.

Casey-Hofland:
Hi Adam, if you see the new commit, I'll just be working on finishing up the stuff I said I want to finish, but if you would find the time to answer my question about Doubles and Quaternions that'd be great!

miikalo:
Hi Casey, thank you for the excellent video which explained this much faster than textual description. I'm a fan of the new Unity Input System and in my project I set up a wrapper script to simply assign the Vector2 from right analog stick to an Atom variable, for example. Would be great to have a more modular system like this in Atoms. I second what Adam already commented in his review, and have an additional question: Since the package depends on the new Unity Input package, can it be described as a dependency in Unity Package Manager somehow? Like if a project wants to add this Atom package, then it would also have to install the new Input System as well to highlight that it won't work without it.

Casey-Hofland:
Yes, this would be very easy to do. Right now it is always installed and ignores all the code if the Input System is not included in your project.
The reason I set it up this way is because of Unity's package manager actually. Unfortunately you can't / they don't make it easy to search for packages online. If you can't find this package online, then I believed it best for it to be included in the bulk integration.
Alas, installing atoms shouldn't mean that you have to install the Input System.
If the package manager was more like nuget I would be 100% on board with this, but due to this limitation I'm just a bit uncertain. But again, as I said: right now it won't do any harm if the Input System is not installed (it won't even be compiled so even editor speed isn't affected.)

AdamRamberg:
> Hi Adam, if you see the new commit, I'll just be working on finishing up the stuff I said I want to finish, but if you would find the time to answer my question about Doubles and Quaternions that'd be great!

What is your question regarding Doubles and Quaternions? :)

Packages/InputSystem/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "files": [
+    "/package.json",
+    "/package.json.meta",
+    "/Runtime",
+    "/Runtime.meta"
+  ],
+  "dependencies": {
+    "com.mambojambostudios.unity-atoms-core": "4.3.0",
+    "com.mambojambostudios.unity-atoms-base-atoms": "4.3.0"
+  }
```
AdamRamberg, line 21: add `"com.unity.inputsystem": "1.0.1"` as a dependency here

Packages/InputSystem/Runtime/Interpreters/CallbackContextInterpreter.cs
```diff
@@ -0,0 +1,82 @@
...
+        where VI : AtomVariableInstancer<V, P, T, E1, E2, F>
+    {
+        [SerializeField] private E1 _started;
+        [SerializeField] private E1 _performed;
+        [SerializeField] private E1 _canceled;
+        [SerializeField] private E1 _waiting;
+        [SerializeField] private E1 _disabled;
+
+        [SerializeField] private V _value;
+        [SerializeField] private BoolVariable _valueAsButton;
```
AdamRamberg, line 24: Change to a `BoolReference`

Casey-Hofland:
Oh sorry, it’s in the commit haha! I need a DoubleVariable and QuaternionVariable (+ interest) to complete the InputSystem and I was wondering if I should put them inside the base atoms package.
This is due to the Input System being able to bind controls to these value types (e.g. gyroscopes/sensors).

AdamRamberg (changes requested):
Additionally we need to add the following to the root `package.json`: 
`"publish:input": "npm publish ./Packages/InputSystem"`
as well as added to the `publish:all` script. 

Instead of manually generate Atoms for `CallbackContext` and `PlayerInput` we should add them to `RegenerateAllAtoms.cs`. For types: Actions, EventInstancers, Events, UnityEvents.

AdamRamberg:
> Hi Casey, thank you for the excellent video which explained this much faster than textual description. I'm a fan of the new Unity Input System and in my project I set up a wrapper script to simply assign the Vector2 from right analog stick to an Atom variable, for example. Would be great to have a more modular system like this in Atoms. I second what Adam already commented in his review, and have an additional question: Since the package depends on the new Unity Input package, can it be described as a dependency in Unity Package Manager somehow? Like if a project wants to add this Atom package, then it would also have to install the new Input System as well to highlight that it won't work without it.
> 
> Great stuff btw!

The `dependencies` in package.json also specifies what it depends on. I'm not 100% sure, but I hope that the package manager takes that into account

AdamRamberg:
> Oh sorry, it’s in the commit haha! I need a DoubleVariable and QuaternionVariable (+ interest) to complete the InputSystem and I was wondering if I should put them inside the base atoms package.
> This is due to the Input System being able to bind controls to these value types (e.g. gyroscopes/sensors).

Aha alright. They should be added to the BaseAtoms package then. The easiest way to do that is to add them to the `RegenerateAllAtoms.cs` script.

Packages/InputSystem/Runtime/Interpreters/CallbackContextInterpreter.cs
Casey-Hofland: That won't really do much... _valueAsButton is only used to set a value that needs to be read from elsewhere, so that crosses out bool or BoolConst. Since CallbackContextInterpreter is a ScriptableObject, that crosses out BoolVariableInstance.

Packages/InputSystem/package.json
Casey-Hofland: Alright, but I think it's fair we make it clear in the ReadMe. Something like: "Optional packages may have additional dependencies."

Casey-Hofland:
New commit, new grid!

I've added Doubles and Quaternions, to RegenerateAllAtoms as well! I couldn't figure out how to use it though (by which I mean the option Tools > Unity Atoms > Regenerate all Atoms). My supposition was that it would regenerate all the scripts in the packages, so I was a bit surprised when it asked me to select a folder. Neither selecting "core" (abandoned the process) or "packages" (copied all atoms instead of replacing them) gave me the desired outcome of

4pw84k

For now I've used the Generator to create the new values and painstakingly dragged them all to their proper location. Am I missing some handy shortcut to all this? I'll still have to make 4 to accommodate for a few special InputControl<T> instances.

Casey-Hofland:
(Don't worry, I'm just pushing in smaller doses, you should be able to knock most of these out in 2 minutes.)

Casey-Hofland:
Hi @AdamRamberg, I was wondering if you've had time to check it out. All the code is ready to go.

Casey-Hofland:
Well... I messed something up trying to get a repo of Packages only. Only the first 8 commits are worth focusing on.

Packages/InputSystem/Runtime/Interpreters/CallbackContextInterpreter.cs
AdamRamberg: Makes sense

AdamRamberg:
No worries regarding the rogue rebase, I will squash this PR anyways. 

I've made a couple of changes to your PR: 
- Reverted some of the changes in the Generator since it caused issues in generation. 
- Added docs
- Fixed asmdef
- Fixed bug in Stack Trace Entry that was found during testing
- Regenerated all atoms which resulted in some new paths for editor scripts for this pkg.

AdamRamberg (approved):

AdamRamberg:
Just merged your PR to canary. This subpackage will be available in the next release of Atoms. Thanks again for the contribution!

Issue #214, opened 2020-12-10
[BUG] (canary) Conditions created do not appear under Atoms Search


miikalo:
**Describe the bug**
Can't create new Conditions using Atoms Search.

**To Reproduce**
1. Create a new Condition with a script from the Conditions tutorial (see PR #206)
2. Use Atoms Search from Create context menu or press Alt+1

**Expected behavior**
Conditions created in scripts are listed as one of the items under the new Atoms Search menu (identical contents to older Unity Atoms context menu under Create).

**Screenshots**
![image](https://user-images.githubusercontent.com/1841625/101828490-e9fc6280-3b31-11eb-82ee-3366a50dc04a.png)

![image](https://user-images.githubusercontent.com/1841625/101829019-a9511900-3b32-11eb-886f-a1909f31cbe3.png)

**Environment**
 - OS: Windows
 - Unity Version: 2020.1.16f1
 - Version of Unity Atoms: canary at c552881

**Additional context**
-

AdamRamberg:
Fixed in canary. Will be included in next release of Unity Atoms. Closing the issue.

PR #216, opened 2020-12-21
Add AtomsSearchable attribute to indentify searchable atoms and apply…

AdamRamberg:
Solves #214. Adds an `AtomsSearchable` attribute to indentify searchable atoms and applies that to `BaseAtom`. Also corrects `EditorIcon`'s usage flag.


soraphis (approved):
looks good 👍

PR #220, opened 2020-12-22
Trigger initial events for instancers

### LINKED ISSUE #210: [BUG] Instance variables do not trigger "Trigger event on OnEnable" -> "Changed"
**Describe the bug**
Creating an instance variable does not trigger "Trigger event on OnEnable" -> "Changed"

**To Reproduce**
-> Create an int variable, tick "Trigger Event on OnEnable" - "Changed"
-> On gameobject create an "Int Variable Instancer" and assign the created int variable
-> Create an "Int Event Reference Listener" and assign the previous variable instancer in "Event Reference"
-> Register an event
-> Click Play
-> The event is not fired OnEnable

**Expected behavior**
Instance variables should trigger "Trigger event on OnEnable" if "Changed" or "ChangedWithHistory" are checked

**Environment**
 - OS: Windows
 - Unity Version 2020.1.9f1
 - Version of Unity Atoms 4.3.0

**Additional context**
Attached demo project with demonstration
[Examples.zip](https://github.com/AdamRamberg/unity-atoms/files/5573419/Examples.zip)

AdamRamberg:
- Trigger initial events for instancers
- Clean up event name for events instantiated at runtime
- Added properties to check if an event is set or not on a variable. This is useful in order to not accidentally create an event at runtime when just checking if it is null or not.
- Added logic to clean up events instantiated at runtime.

Fixes #210

### DIFF
--- a/Packages/Core/Editor/Editors/AtomGeneratorEditor.cs
+++ b/Packages/Core/Editor/Editors/AtomGeneratorEditor.cs
@@ -30,7 +30,6 @@ namespace UnityAtoms.Editor
 
         public override void OnInspectorGUI()
         {
-            Rect buttonRect = new Rect();
             var rect = GUILayoutUtility.GetRect(new GUIContent("Show"), EditorStyles.toolbarButton);
 
 
--- a/Packages/Core/Runtime/EventReferences/AtomEventReference.cs
+++ b/Packages/Core/Runtime/EventReferences/AtomEventReference.cs
@@ -12,9 +12,9 @@ namespace UnityAtoms
     /// <typeparam name="VI">Variable Instancer of type `T`.</typeparam>
     /// <typeparam name="EI">Event Instancer of type `T`.</typeparam>
     public abstract class AtomEventReference<T, V, E, VI, EI> : AtomBaseEventReference<T, E, EI>, IGetEvent, ISetEvent
-        where V : IGetEvent, ISetEvent
+        where V : IGetOrCreateEvent, ISetEvent
         where E : AtomEvent<T>
-        where VI : IGetEvent, ISetEvent
+        where VI : IGetOrCreateEvent, ISetEvent
         where EI : AtomEventInstancer<T, E>
     {
         /// <summary>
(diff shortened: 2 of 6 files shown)


soraphis:
> Added properties to check if an event is set or not on a variable. This is useful in order to not accidentally create an event at runtime when just checking if it is null or not.

i think this is an important change, but it should go even further:

- it is quite natural to do `variable == null` so we cannot prevent other developers from doing `x.Changed == null` which could result in the same error prone situation.
- the Changed property should actually be removed or reduced to a naiive property.
- Instead a 'GetOrCreate' method should be used

AdamRamberg:
> > Added properties to check if an event is set or not on a variable. This is useful in order to not accidentally create an event at runtime when just checking if it is null or not.
> 
> i think this is an important change, but it should go even further:
> 
>     * it is quite natural to do `variable == null` so we cannot prevent other developers from doing `x.Changed == null` which could result in the same error prone situation.
> 
>     * the Changed property should actually be removed or reduced to a naiive property.
> 
>     * Instead a 'GetOrCreate' method should be used

Sounds good to me! However, we need to make sure that listeners are using `GetOrCreate` when retrieving the event., since I believe that the auto creation of events at runtime got added so that you don't need to add an event to a simple variable that is just going to be listened to in one listener (read simplify the most basic workflow). In order to make sure listeners are using `GetOrCreate` I believe we need to use it in variables' and variable instancers' `GetEvent` methods.

miikalo (approved):
I tested the branch and it works like a charm. The private boolean variables have a "code smell" about them, but I can't quite articulate how to improve it.

Casey-Hofland:
Now I am no veteran in Atoms by any means but if I may give my 2 cents: I don't think it is okay to generate objects at runtime that just aren't set, because they aren't set for a reason. If Changed is null, it 'should' throw an error (GetOrCreate of course kinda fixes this.) However, if you want to go that road, why not make AtomVariable inherit from AtomEvent? Currently a bit of a stretch, but that should be totally viable for V5 with generic serialization support and all that. Just imagine being able to drag in a PlayerSpeed object in a listener: massive object reduction.

AdamRamberg:
@Casey-Hofland I've been thinking about that back and forth before and that is definitely something to think about for v5. For this though PR I will go ahead and add GetOrCreate for Variables / Instancers and then use that in Listeners. The only thing that will differ from last release then, is that checking the `Changed` event won't auto create anymore. Users that are depending on that behaviour will need to migrate there code to use `GetOrCreate` instead. I believe that should be acceptable for a new minor release.

AdamRamberg:
Added a `GetOrCreateEvent` as discussed. @soraphis can you take a second look?

soraphis (approved):

Issue #194, opened 2020-09-23
[FEATURE] Replay Buffer Settings


tivomk:
**Is your feature request related to a problem? Please describe.**
The Replay Buffer on Events is not always a wanted feature. It can get very tedious to disable it manually on every new event and to make sure no event has it on "by accident" which can lead to sometimes hard to recognize bugs.

**Describe the solution you'd like**
It would be great if there was an option to disable the Replay Buffer on Events. Either by giving the event.Register(Action) method an optional flag to disable the replay buffer for this listener or by generally disabling it on an event asset.

**Describe alternatives you've considered**
Alternatively an option to set a global default value for the Buffer would be an acceptable solution though the safer option would be to have freedom over which listeners want to use the buffer and which do not want to use the buffer.

TowerCrow:
Disabling this feature is the first thing I have to do on every new event. Something that is optional and unnecessary most of the time should not be a setting that's enabled by default.

AdamRamberg:
Thanks for reporting this issue! 

The reason for the Replay Buffer having an initial value of 1 is because a timing issue that will occur if the Replay Buffer is set to 0. Since 1) Variables are triggering there "initial value events" on OnEnable and 2) Listeners are subscribing to Events on OnEnable, it is not certain that Listener's OnEnable will run before Variable's OnEnable. This results in that sometimes Listeners' handlers will run and sometimes not. The current solution (which might not always be preferable) has been the least complex solution to this problem. 

One solution to the issue that I'm describing above might be to use the attribute `DefaultExecutionOrder` for Listeners and Variables. I'm guessing that there is some issues regarding this, since it wasn't done from the beginning, but it's never the less worth investigating again. 

If we can solve the issue that I'm describing above in an elegant way, then I'm fine with the Replay Buffer being initialized to 0. However, I believe that in that case it would be preferable to be able to globally configure Atoms to use another default Replay Buffer value (using user prefs). 

@tivomk Can you please explain your desired solution further? I don't see how your solution is less tedious then manually setting the Replay Buffer to 0 for each newly created Event. Setting the Replay Buffer to 0 on an Event is the same thing as disabling the Replay Buffer for that Event.

tivomk:
The resaon why I would like the solution in code is, that this would ensure the behaviour no matter what is set in the inspector for that particular listener.

[...]

PR #225, opened 2020-12-30
#194 Basic solution to fix replay buffer settings set to 1

ThimoDEV:
Addresses #194 

Sets a default execution order attribute for all listeners and variables so that Listeners OnEnable() runs before Variables OnEnable().

This should allow new created event replay buffer settings be initialized to 0.

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -31,7 +31,7 @@ namespace UnityAtoms
         [SerializeField]
         [Range(0, 10)]
         [Tooltip("The number of old values (between 0-10) being replayed when someone subscribes to this Event.")]
-        private int _replayBufferSize = 1;
+        private int _replayBufferSize = 0;
 
         private Queue<T> _replayBuffer = new Queue<T>();
 
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -14,6 +14,9 @@ namespace UnityAtoms
     {
         public T InspectorRaiseValue { get => _inspectorRaiseValue; }
 
+        [SerializeField]
+        private bool UseLocalReplayBuffer = false;
+
         /// <summary>
         /// Retrieve Replay Buffer as a List. This call will allocate memory so use sparsely.
         /// </summary>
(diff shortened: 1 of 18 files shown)


AdamRamberg:
Thanks @ThimoDEV for submitting this PR! 🙏

We need to do 2 things before approving. First, we need to address Variable Instancers in this PR and set their DefaultExecutionOrder. Secondly, we should also add a preference field (AtomsPreferences.cs) which specifies default buffer size on events.

ThimoDEV:
@AdamRamberg 

Changed the default execution order code because Runtime.cs had already constants for the variable instancers and I added the variable and listener ones there. Added a new Atoms setting to set a global replay buffer value. For now it doesnt check the set value. Any best suggestion on how to do this. I've been thinking about the following:

Add a regex check to restrict user input.

ThimoDEV:
@AdamRamberg 

Made changes based on your suggestions, there is now a global slider for the replay buffer. The only thing that is missing is a label at the end of the slider to show the current value. You know how I can add that?

Packages/Core/Runtime/Events/AtomEvent.cs
```diff
@@ -31,10 +31,15 @@ public class AtomEvent<T> : AtomEventBase
...
         [Range(0, 10)]
         [Tooltip("The number of old values (between 0-10) being replayed when someone subscribes to this Event.")]
-        private int _replayBufferSize = 1;
+        private int _replayBufferSize;
 
         private Queue<T> _replayBuffer = new Queue<T>();
 
+        private void OnEnable()
+        {
+            _replayBufferSize = AtomPreferences.ReplayBufferSize;
```
AdamRamberg, line 40: This will just overwrite the replay buffer size if set from the Inspector. Not 100% sure how to do this properly though.

AdamRamberg (changes requested):
This PR breaks the InfiniteWaves demo, since Listeners are running their OnEnable before Variable / Event Instancers are running. This causes errors when Listeners are trying to retrieve Events from the Instancers. Listeners should be executed after all others, not sure though if that might have other consequences. 

For this to be merged I need 2 changes: 
- Fix the assignment of the default preference value.
- InfiniteWaves demo can't break because of this change. 

About your question regarding the slider. Not sure about getting the value as a label next to the slider. I think however that the slider is good enough as it is.

miikalo:
The InfiniteWaves demo produces no errors, but it also does not spawn any enemies.

ThimoDEV:
My apologies, I changed the events by setting the use local boolean and set the replay buffer to their original value. I see now these event files conflict, but I cant see what their conflict is, or how should I do that?

Casey-Hofland:
Since this branch is broken and outdated, I will be closing it, as was discussed in #263 and #264.

PR #226, opened 2020-12-30
Generic property drawers

Casey-Hofland:
I wanted to create this draft so I could explain the plan, because it is pretty involved. And yes: I should definitely have used my microphone for this and the 10mb limit is weighing me down.

https://user-images.githubusercontent.com/27729987/103387132-6ddab680-4b02-11eb-8334-67be17a354fc.mp4

It's a big change that I think is crucial to get out early so the rest of V5 development can benefit from it.


Casey-Hofland:
Gosh, this is a big one! I also wanted to remove the generated editors, however this PR is pretty involved already so I thought I'd first gauge your reaction to this one before moving on.

AdamRamberg:
@Casey-Hofland This is really promising! I don't see any issue using reflection in editor scripts so no worries when it comes to that. Shaving off generated code like this is really the number 1 goal of v5 so that is also good. However, I'm not 100% sure that we should continue on this branch or not, a lot of changes have been added to master since I made this branch. I think we should discuss this internally among the maintainers before we make a decision on how to proceed.

Casey-Hofland:
What version is this branch from? Maybe all it needs is a good sync (which is hopefully as easily done as said). We can try it on a forked branch so there’s no risk.

It is, however, a strange conundrum, even looking past the current issue. Once V5 becomes the new master, how is V4 going to keep receiving support in a manner that doesn’t involve everyone writing the same code twice?

PR #227, opened 2021-01-01
Fix the Discord link in the FAQ

matthewtole:
Noticed that the Discord invite link in the FAQ was broken, so updating it to match the one in the footer.

_Not sure if you need a Github issue opened for a PR this small, happy to make one if required._


AdamRamberg:
Good catch! No need for an issue :)

PR #230, opened 2021-01-02
compatibility with dynamic assemblies and different namespace/class conditions

### LINKED ISSUE #231: [BUG] Dynamic Assemblies like URP make asset based generator to throw exceptions [BUG]
**Describe the bug**
When the Asset based generator's type dropdown makes an lookup for serialized types it throws exceptions on dynamic assemblies. the URP assembly seems to be dynamic

**To Reproduce**
add Univseral RP via package manage, try to select a type in the asset based generator

soraphis:
fixes #231 #232 #233

### DIFF
--- a/Packages/Core/Editor/Editors/AtomFolderEditor.cs
+++ b/Packages/Core/Editor/Editors/AtomFolderEditor.cs
@@ -1,4 +1,4 @@
-﻿using UnityEngine;
+using UnityEngine;
 using UnityEditor;
 using System.IO;
 
@@ -23,6 +23,7 @@ namespace UnityAtoms.Editor
                     AssetDatabase.GetAssetPath(target),
                     "New Atom Type.asset"
                 );
+                newPath = AssetDatabase.GenerateUniqueAssetPath(newPath);
                 Debug.Log(newPath);
                 AssetDatabase.CreateAsset(asset, newPath);
                 AssetDatabase.ImportAsset(newPath);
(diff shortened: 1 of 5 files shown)


Casey-Hofland:
The fix didn't work for me, but I've found the issue (it's a doozy). URP has a class Readme, with a nested class Section. This type's Namespace is null, causing the generators query to fail.

Ignoring anything with a null namespace is not an option, as the InputSystem package relies on nested types being supported by the generator.

Casey-Hofland (changes requested):
This prevents the generator from crashing, but generating a nested types doesn't work.

e.g.
[Serializable] public class Foo
{
    [Serializable] public class Bar {}
}

will generate
public sealed class BarConstant : AtomBaseVariable<Bar> (should be BarConstant : AtomBaseVariable<Foo.Bar>)

BarConstant should maybe get a different name, as it might conflict with existing generated classes. Foo_BarConstant would be the safest, although slightly awkward. We can discuss this later.

Please test generating a nested type and request a review upon a test without any errors.

soraphis:
corrected an error in my previous fix for no-namespace classes.
corrected generation for nested classes

Casey-Hofland (changes requested):
Generator not working

Can you please test generating Again with this script and ask for a review only after you can successfully generate it without any errors?
[Something.txt](https://github.com/AdamRamberg/unity-atoms/files/5761752/Something.txt)

Casey-Hofland (commented):
Please read my commit messages and ask me questions if they are uninterpretable (some I wrote kinda quick).

miikalo (approved):

PR #238, opened 2021-01-10
Lazy stacktrace toString conversion

soraphis:
when disabling atoms debug mode, StackTraceEntries are allocated, but their strings are never used.

this costs much performance in the InfinitiveWave Demo.
create many enemies and make them move. observe the unity profiler


miikalo (approved):

Casey-Hofland:
The InfiniteWaves Demo is resetting waves to 1 when I play and I can't find where to set it. It also keeps giving me a NullReferenceException that my Player has 2 Tags, one of which is null, even though I see this:
Screenshot 2021-01-12 001318

It makes it hard to test this PR, but if you like it @miikalo I trust it works as advertised.

Casey-Hofland (approved):

Issue #240, opened 2021-01-16
[BUG] Vector3Reference.Value throws NullReferenceException in OnDrawGizmos while not in Play mode


asasine:
**Describe the bug**
Using a `Vector3Reference` assigned to a `Vector3 Variable Instancer` in the `OnDrawGizmos` and `OnDrawGizmosSelected` object while not in Play mode causes a `NullReferenceException` to be thrown every few seconds. This occurs even if access to the `Vector3Reference.Value` property is guarded behind a call to `vector3Reference.IsUnassigned`.

This error does not occur when the `Vector3Reference` is assigned to a Value, Constant, or a Variable. It does not occur in Play mode when the reference is assigned to an instancer.

**To Reproduce**
1. Exit play mode.
1. Create a `Vector3 Variable Instancer` on a GameObject
1. Create a `Vector3Reference vector3Reference` field on a `MonoBehavior`
1. Assign the `Vector3 Variable Instancer` to your new `vector3Reference` field in the inspector
1. Add the following code to your `MonoBehaviour`

    ```csharp
    private void OnDrawGizmos()
    {
        if (!this.vector3Reference.IsUnassigned)
        {
            Gizmos.DrawSphere(this.vector3Reference.Value, 1f);
        }
    }
    ```

1. Watch a new `NullReferenceException` appear every few seconds in the Console
1. Enter Play mode
1. Watch exceptions stop

**Expected behavior**
Accessing the `Vector3Reference.Value` should be safe if guarded first by a call to `!vector3Reference.IsUnassigned` or another similar property.

**Screenshots**
![image](https://user-images.githubusercontent.com/5421969/104809141-118dbd00-57b9-11eb-939a-2fe465e15bd7.png)

**Environment**
 - OS: Windows 10 version 1909
 - Unity Version: 2020.2.0f1
 - Version of Unity Atoms: 4.4.2

**Additional context**
Stack trace:
```
NullReferenceException: Object reference not set to an instance of an object
UnityAtoms.AtomBaseVariableInstancer`2[T,V].get_Value () (at Library/PackageCache/com.mambojambostudios.unity-atoms-core@4.4.2/Runtime/VariableInstancers/AtomBaseVariableInstancer.cs:31)
UnityAtoms.AtomReference`8[T,P,C,V,E1,E2,F,VI].get_Value () (at Library/PackageCache/com.mambojambostudios.unity-atoms-core@4.4.2/Runtime/References/AtomReference.cs:38)
[...]

PR #241, opened 2021-01-16
fixing #240 - Value of a VariableInstancer in EditMode leads to NullRef

soraphis:
See #240 
should fix a nullref exception when checking the Value of a VariableInstancer on a Reference might lead to null reference exceptions


Casey-Hofland (approved):
Great stuff! I've only added a very small to change to AtomBaseVariableInstancer (Start is called only once, which is safe in this case).

PR #242, opened 2021-01-16
Generic Editors

Casey-Hofland:
Yeeaah, I know we were going to wait with this one but I was kinda wondering what would need to change in order for the editors to work. Not that much I found (though getting the Reflection to work was a small hassle), so I figured I'd go ahead and implement it. A bit premature, but the water is fine.


bguyl:
Maybe it's the good time to make VariableEditor use the UI Toolkit as well

Casey-Hofland:
We are considering this internally, but only for V5.1 at the soonest. It's also the question if we want to invest time in refactoring that code when there is no real need.

bguyl:
I was thinking at, at least, having a `VisualElement` at root and maybe keeping the actual behavior in a `IMGUIContainer`
This way, Variable and Event editor will share approximately the same structure

Issue #243, opened 2021-01-17
[BUG] Variable is sometime resetting between scenes in build


bguyl:
**Describe the bug**
In build, Unity unloads unused assets when switching scenes, so sometimes AtomsVariable are reseting, sometimes not.
Example:
SceneA is referencing a FloatVariable
SceneB is referencing the same FloatVariable
SceneC is not referencing the Variable.

So if you go A -> B, the asset is in used, Unity don't unload it, the value is kept
If you go A -> C, the asset is considered unused, Unity unload it

**To Reproduce**
Reproduction repository: https://github.com/bguyl/unity-so-bug
Build download: https://github.com/bguyl/unity-so-bug/releases/tag/0.0.1

**Expected behavior**
Variable and event behaviors should be consistent between editor play and build.
Either values should reset at each scene change or the should not reset at all.

**Screenshots**
![so-bug](https://user-images.githubusercontent.com/24865257/104845299-eb5d3f80-58d4-11eb-9911-01c3445ac869.gif)

**Environment**
 - Unity Version: 2020.2.1
 - Version of Unity Atoms: 4.4.2

**Additional context**
If you mark SOs as Addressable, the issue is fixed as the asset lifetime is manually managed by you.
Two others way to enforce consistency behavior is:
- Mark SOs as [DontUnloadUnusedAsset](https://docs.unity3d.com/ScriptReference/HideFlags.DontUnloadUnusedAsset.html) to make them act like a "Global" variable
- Reset values on [ActiveSceneChanged](https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager-activeSceneChanged.html)

fakegood:
Hello, i would like to know is your scene marked as Addressables?
and generally is Addressables installed in your project?

bguyl:
👋
No, in the reproduction project, addressable is not installed.
When I talk about addressable that's because I use the package in another project.
Do you have something in mind?

fakegood:
So from what I know, scene will be converted to another form of addressable even if you don't mark it as one.

When addressable is built, if a scriptable object is referenced in the scene, it will grab that asset and pack it together.

If there's 2 scenes referencing a scriptable object, each scenes will have an independent reference of the asset, because it sad never know that those assets referenced are the same assets.

That is why when you mark the scriptable object as an addressable, the referenced asset is maintained between scenes.

[...]

PR #248, opened 2021-01-29
#243 Add scope (or lifetime) management

bguyl:
Hi :wave:

This is WIP, but I wanted to share my work and maybe open the discussion

### Runtime changes

- Add a `Scope` Enum
- Add a `ManageLifetime` method in `BaseAtom` as it could be used by Variables, Lists and Collections
- In `AtomVariable`, when scriptable object is enabled its behavior will be defined
    * `Scope.Global`: Will set the [DontUnloadUnusedAsset](https://docs.unity3d.com/ScriptReference/HideFlags.DontUnloadUnusedAsset.html) hide flag, the SO will not be unload automatically during runtime
    * `Scope.Scene`: Will add a listener to [SceneManager.activeSceneChanged](https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager-activeSceneChanged.html) that will reset the SO value to `InitialValue`
- FSM script will call base `OnEnable`
  
### Editor changes

- Define #USE_ADDRESSABLES if addressables package is installed
- Show an HelpBox in Variable and Event inspector is addressables is installed or if the asset is marked as addressable
- In variable editor, force the scope value to `Unmanaged` is the asset is marked as addressable
- In variable editor, disable scope edition in runtime

### What is missing

- Documentation about lifetime, why it's important and how to work with addressables
- List and Collection lifetime management (and their editors). I might want to wait for generic editors for this

Other atoms type such as constants or actions should be ok with this issue

### DIFF
--- a/Packages/Core/Editor/Editors/Events/AtomEventEditor.cs
+++ b/Packages/Core/Editor/Editors/Events/AtomEventEditor.cs
@@ -1,6 +1,11 @@
 #if UNITY_2019_1_OR_NEWER
 using UnityEngine;
 using UnityEngine.UIElements;
+using UnityEditor;
+#if USE_ADDRESSABLES
+using UnityEditor.AddressableAssets;
+using UnityEditor.AddressableAssets.Settings;
+#endif
 
 namespace UnityAtoms.Editor
 {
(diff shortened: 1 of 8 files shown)


bguyl:
I already posted this on Discord, but here is some showcase of my early work:

scope-addressable
scope-test

miikalo:
Hi @bguyl and thanks for the PR! Unfortunately it can't target the canary branch, as that is intended for v4 bugfixes. Sorry if this isn't clearly communicated, we're working on a "developer guide" for that purpose. 

I'm changing the target branch to v5, which is our next major release. However, that branch might be in a broken state for a while as plenty of major changes will take place. You can either adapt to v5 or wait until v5 is a bit more stable, the PR content itself has merit and will be considered for v5.

miikalo:
Looks like this could use a rebase or something. Lifetime has been a topic in some discussions on Discord and I think it would be useful to be explicit about it.

bguyl:
I tried to create another branch from v5 and just cherry-picking my commit, but the API changed too much, some modifications didn't go so well
I think I will redo my changes from v5, it will be safer

PR #263, opened 2021-04-04
Pull master into V5

Casey-Hofland:
This will pull the master branch into V5 so that development on V5 can go into full effect.

For anyone reading this: don't get too excited. We don't have an ETA and unfortunately the team is in a very busy period, everyone for their own reasons.

The only thing left to pull is an outdated request which I'd personally like to close: let me know if there are any votes against this.
https://github.com/unity-atoms/unity-atoms/pull/225

Since this is a very important step, I've requested all of you to approve this merge before I go through with it. After that, I can try to add one or two features to V5 once in a while (but no promises!)

And also: it's good to be back, if even for a moment ^_^


Casey-Hofland:
I'm going to create a copy of the master branch to merge into v5, because some things from the master branch need to be deleted in order for a smooth merge process.

PR #266, opened 2021-04-13
Fixes #265 Atom event code cleanup

### LINKED ISSUE #265: [FEATURE] Code Cleanup
**Is your feature request related to a problem? Please describe.**
I noticed a few areas in the Core package where some code could be a little cleaner. Specifically in AtomEvent and AtomEventBase.

**Describe the solution you'd like**
I've tidied up a few areas that I thought had some room for improvement. I'm creating this issue so it's available to link against an incoming PR.

schodemeiss:
Fixes #265. Whilst I was working on some utils for Unity-Atoms today and did a little [Boy Scouting](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) whilst in the process.

Most of the focus was around AtomEvent and AtomEventBase as those had some glaring "noise" in them. Overall, I did the following:

- Fixed typos.
- Added doc blocks for every method and property.
- Simplified clearing all the delegates when exiting play mode.
- Reduced nesting in some longer methods.
- Made a handful of private properties _readonly_ where applicable.
- Avoid null check against value type in IListExtensions.First().
- Throw InvalidOperationException in AtomFunction.cs as per .NET guidelines (https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types).
- Use `foreach` in VariableResetter.cs. `foreach` doesn't generate garbage when iterating over List<> since Unity 5.6.
- Reduced nesting in Observables.

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -6,23 +6,30 @@ using UnityEngine;
 namespace UnityAtoms
 {
     /// <summary>
-    /// Generic base class for Events. Inherits from `AtomEventBase`.
+    /// Generic base class for Events.
     /// </summary>
     /// <typeparam name="T">The type for this Event.</typeparam>
     [EditorIcon("atom-icon-cherry")]
     public class AtomEvent<T> : AtomEventBase
     {
-        public T InspectorRaiseValue { get => _inspectorRaiseValue; }
+        /// <summary>
+        /// Value used when raising events at runtime with the inspector.
+        /// </summary>
+        public T InspectorRaiseValue => _inspectorRaiseValue;
 
         /// <summary>
         /// Retrieve Replay Buffer as a List. This call will allocate memory so use sparsely.
         /// </summary>
-        /// <returns></returns>
-        public List<T> ReplayBuffer { get => _replayBuffer.ToList(); }
+        public List<T> ReplayBuffer => _replayBuffer.ToList();
 
+        /// <summary>
+        /// How many items to store in the replay buffer.
+        /// </summary>
         public int ReplayBufferSize { get => _replayBufferSize; set => _replayBufferSize = value; }
 
-        [SerializeField]
+        /// <summary>
+        /// <see cref="MulticastDelegate"/> to invoke when the event is raised.
+        /// </summary>
         protected event Action<T> _onEvent;
 
         /// <summary>
@@ -33,19 +40,17 @@ namespace UnityAtoms
         [Tooltip("The number of old values (between 0-10) being replayed when someone subscribes to this Event.")]
         private int _replayBufferSize = 1;
 
-        private Queue<T> _replayBuffer = new Queue<T>();
+        /// <summary>
+        /// Queue used as a replay buffer.
+        /// </summary>
+        private readonly Queue<T> _replayBuffer = new Queue<T>();
 
+        /// <summary>
+        /// This function is called when the scriptable object goes out of scope.
+        /// </summary>
         private void OnDisable()
         {
-            // Clear all delegates when exiting play mode
-            if (_onEvent != null)
-            {
-                var invocationList = _onEvent.GetInvocationList();
-                foreach (var d in invocationList)
-                {
-                    _onEvent -= (Action<T>)d;
-                }
-            }
+            _onEvent = null;
         }
 
         /// <summary>
@@ -70,9 +75,8 @@ namespace UnityAtoms
         }
 
         /// <summary>
-        /// Used in editor scipts since Raise is ambigious when using reflection to get method.
+        /// Used in editor scripts since Raise is ambiguous when using reflection to get method.
         /// </summary>
-        /// <param name="item"></param>
         public void RaiseEditor(T item) => Raise(item);
 
         /// <summary>
@@ -86,7 +90,7 @@ namespace UnityAtoms
         }
 
         /// <summary>
-        /// Unregister handler that was registered using the `Register` method.
+        /// Unregister handler that was registered using the <see cref="Register"/> method.
         /// </summary>
         /// <param name="action">The handler.</param>
         public void Unregister(Action<T> action)
@@ -95,7 +99,7 @@ namespace UnityAtoms
         }
 
         /// <summary>
-        /// Unregister all handlers that were registered using the `Register` method.
+        /// Unregister all handlers that were registered using the <see cref="Register"/> method.
         /// </summary>
         public override void UnregisterAll()
         {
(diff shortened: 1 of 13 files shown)


Packages/Core/Runtime/Events/AtomEvent.cs
```diff
@@ -33,19 +40,17 @@ public class AtomEvent<T> : AtomEventBase
         [Tooltip("The number of old values (between 0-10) being replayed when someone subscribes to this Event.")]
         private int _replayBufferSize = 1;
 
-        private Queue<T> _replayBuffer = new Queue<T>();
+        /// <summary>
+        /// Queue used as a replay buffer.
+        /// </summary>
+        private readonly Queue<T> _replayBuffer = new Queue<T>();
```
soraphis, line 46: reminds me of: https://i.redd.it/iuy9fxt300811.png

is a docstring for the sake of docstring actually necessary? especially on a _private_ field? what have you won? it is of Type Queue and its name is replayBuffer, so all the comment says is "used as a"

Packages/Core/Runtime/Events/AtomEvent.cs
```diff
@@ -33,19 +40,17 @@ public class AtomEvent<T> : AtomEventBase
...
 
+        /// <summary>
+        /// This function is called when the scriptable object goes out of scope.
+        /// </summary>
```
soraphis, line 50: is it really possible to repeat the unity documentation?

https://docs.unity3d.com/ScriptReference/ScriptableObject.OnDisable.html

here again, this is not a public interface, (which SHOULD be described) but instead a private method, within the context of unity. basically every maintainer should know by hearth when OnDisabled is called

Packages/Core/Runtime/Events/AtomEvent.cs
```diff
@@ -104,63 +108,68 @@ public override void UnregisterAll()
...
+        private void TrimReplayBuffer()
         {
-            if (_replayBufferSize > 0)
-            {
-                while (_replayBuffer.Count >= _replayBufferSize) { _replayBuffer.Dequeue(); }
-                _replayBuffer.Enqueue(item);
-            }
+            while (_replayBuffer.Count >= _replayBufferSize)
+                _replayBuffer.Dequeue();
         }
```
soraphis, line 154: this change does nothing for readability as i now have to scan 10+ LOC more, it does nothing for DRY (since the method is only used at this one place).
furthermore: when introducing new methods, one must carry about not putting them between those two, since it would even more harm readability as i know have to scroll.

A maybe better change to this would be creating a RingBuffer class, which cares about keeping the size constant. that would be a favor in terms of separation of concerns.

It may introduce a (probably negligible) performance impact - hopefully the compiler will inline this. (a local function would always be inlined)

Packages/Core/Runtime/Events/AtomEvent.cs
```diff
@@ -104,63 +108,68 @@ public override void UnregisterAll()
...
 
+        /// <summary>
+        /// Replay buffer into a subscriber.
```
soraphis, line 157: this is the name of the method ... it does not explain anything.
Wouldn't the more obvious summary be: "Invokes the Action `action` on every element in the replayBuffer" 

btw ... wouldn't the code look cleaner with: `using(var enumerator = ... `
or in other words ... why isn't a foreach loop used here?

this method body could be 3 lines of code

Packages/Core/Runtime/Events/AtomEvent.cs
schodemeiss: Very true, just "looked" odd to me where most other parts of the class had docstrings and that didn't. Happy to remove if you feel it's adding to cognitive load.

Packages/Core/Runtime/Events/AtomEvent.cs
schodemeiss: I'm in the habit of not assuming not everyone is going to be 100% familiar with all the Unity callbacks. Whilst I agree all the maintainers _should_ know this is a Unity callback, I didn't want to assume that was the case. Again, happy to remove.

Packages/Core/Runtime/Events/AtomEvent.cs
schodemeiss: Sure, I'll roll this out.

I was considering adding another class similar to your RingBuffer suggestion, but that felt a little over engineered. I'm not actually not super keen on the fact that the event is handling it's own replay buffer at all. Feels like that should behaviour should be "pluggable" per Event, but that's for another day :-).

Packages/Core/Runtime/Events/AtomEvent.cs
schodemeiss: I'll change the comment.

I didn't actually look much into what the method was doing TBH. I don't use the replay buffers for anything, so I didn't look at changing it to "using". But I can roll that into a commit.

Packages/Core/Runtime/Extensions/IListExtensions.cs
```diff
@@ -9,17 +10,15 @@ internal static class IListExtensions
...
 
             return default(T);
         }
 
         public static bool Exists<T>(this IList<T> list, Func<T, bool> func)
         {
-            var first = list.First<T>(func);
-            return first != null ? true : false;
+            var first = list.First(func);
+            return first is { };
```
soraphis, line 21: I'm not entirely sold on this one.
1. this is a pretty modern syntax (using an empty property pattern)
2. its super hard to google if one does not know what it is.
3. its less readable than !(first is null)
4. `first != null` uses the operator overloading for `!=`. `is` does not do that, though. In terms of unity this makes quite a difference when T is a UnityEngine.Object which have an overload for `!=`.
5. First<T> as it was previous uses the extension method above, using Linq.First on the other hand will **throw an exception if no match was found** - there is a FirstOrDefault, exactly doing what the First<T> overload does...

soraphis (changes requested):

soraphis:
I don't know about removing brackets for one-statement loops / if-blocks.
I thought we had a code convention for that case, but could not find it that fast. 

I requested a change in IListExtensions.cs the other things I commented are just that: comments. 
Hopefully someone else can also share their opinion on those cases.

Packages/Core/Runtime/Extensions/IListExtensions.cs
schodemeiss: My reason for changing it was due to the possibility of performing a null check on a value type, but if you think !(first is null) is more readable, I can swap it out to that.

**Note:** From what I can tell, the entire Exists method isn't actually used anywhere. So I think the correct solution is to remove the code entirely.

schodemeiss:
> I thought we had a code convention for that case, but could not find it on a quick search
I used the .editorconfig provided and it didn't "fix" them. There didn't seem to be a consistent approach in the codebase, so I went for the leanest amount of braces.

> I requested a change in IListExtensions.cs the other things I commented are just that: comments.

Thank-you very much for your time & feedback. Always helpful. I've made 2 commits to address some of the things you've already mentioned (especially around the GetEnumerator, I never looked at what that was actually doing, but that should be neater now). I look forward to further feedback from yourself and / or other maintainers.

Casey-Hofland (changes requested):
For starters: your summaries are great, and I'll surely keep them in my back off my head once we're rewriting this code.

However, that's just it. I'm terribly sorry but it seems you have been a bit misfortunate in picking the V5 branch for code cleanup. This branch is currently still in beta and a lot of it will still be rewritten and removed to make use of Unity's ability to serialize generics. To that end, I can't approve this PR until most of the heavy lifting is finished.

If I had to point out any changes, is that the readonly modifiers should go since these may introduce breaking changes to some projects (if they do I would be surprised, but it's not for us to decide). Though I wouldn't, uhm, waste your time on it. Sorry that sounds really bad, it's just kind of the way V5 is right now.

schodemeiss:
> However, that's just it. I'm terribly sorry but it seems you have been a bit misfortunate in picking the V5 branch for code cleanup. This branch is currently still in beta and a lot of it will still be rewritten and removed to make use of Unity's ability to serialize generics. To that end, I can't approve this PR until most of the heavy lifting is finished.

That's precisely why I went for the v5 branch. I figured it was under active development but didn't have a release yet. I figured being a major version, even if there were some public facing changes (of which I don't think there are many in the PR besides the Exception type changes) that wouldn't be a huge deal. But I'm happy to close the PR. Only took a few minutes to put together.

Regarding the `readonly` modifiers, I think only one of them is marked as "protected", so that one _could_ be being used by someone, but the others were all private, so I figured they are just implementation details which no-one could be using in the first place (otherwise, they'd be public or protected).

Issue #269, opened 2021-04-25
[BUG] ChangedOnOnEnable not triggered


simonbucher:
Hi 

Thanks a lot for creating this fantastic framework. It’s a joy to use and helps to keep a decoupled architecture.

I have the following issue.

I’ve created an atom Score for an IntVariable. I use an IntEventRefenceListener attached to a GameObject to Update the Score text. I’ve connected the Listener to the variable with the option „Use Variable “. In the Atom of the Variable, I’ve selected the option „Tigger event on enable“. 

In the editor, this works as I expected. On game Start, the variable gets triggered and updates the UI with the Initial value. If I run this on Android, the event doesn’t get triggered at the game start. All the following events get triggered. Just this first „Trigger Event on Start“ is missing. Enabling „Replay Event Buffer On Register“ doesn’t help in this case.

<img width="530" alt="Bildschirmfoto 2021-04-25 um 12 43 07" src="https://user-images.githubusercontent.com/46404021/115998583-3085c180-a5e8-11eb-950a-cffe34b5f7ab.png">
<img width="530" alt="Bildschirmfoto 2021-04-25 um 12 45 27" src="https://user-images.githubusercontent.com/46404021/115998584-31b6ee80-a5e8-11eb-9d95-6bf8c018abf8.png">

simonbucher:
It looks like, when using an Event Reference Listener with a variable, GetOrCreateEvent() is sometimes called after TriggerInitialEvents(). I will create a PR with a possible solution.

PR #270, opened 2021-04-25
#269 Fix ChangedOnOnEnable not triggered

simonbucher:
This commit is a possible solution for #269

When using an Event Reference Listener with a variable, GetOrCreateEvent() is sometimes called after TriggerInitialEvents(). This commit is a way to solve this.


miikalo (approved):
Tried it, fixes the bug as described. Using a variable indeed does not have the same effect as using an event in this case.

miikalo:
Merging this, it's been around for a while.

Issue #235, opened 2021-01-03
[FEATURE] Utilize ScriptableObject AtomGenerator in Packages


CaseyHofland:
The SO AtomGenerator is a very clever idea and one that I think we should utilize inside of the subpackages. 

I also have a couple of personal wishes for it, which can be implemented separately.

1. Menu restructuring.
Currently, the structure is:
`Unity Atoms > Atom Type(s) > Type`
for example:
`Unity Atoms > Constants > Float`

I would like to use
`Unity Atoms > Namespace* > Type > Atom Type` * = one level deeper for every namespace
for example:
`Unity Atoms > Base Atoms > Float > Constant`

This would more naturally identify a file's location and could help developers understand the effects of removing a package.

2. Folder restructuring.
Currently, when an atom is generated, its files are split into subfolders for Actions, Variables, Constants and many more. I would rather have all the files under a single folder {type}Atom (for example FloatAtom), so it's easier to delete/move around if something went wrong with the generation or if someone wants to reorganize their project.

3. ~ make the Generator partial?
Perhaps if we make the Generator a partial class, packages can extend it? Not sure which side of genius vs madness that falls on, but it would be interesting to explore.

soraphis:
Closing v5 PRs as there is no active development on v5 anymore

Issue #61, opened 2019-10-16
Create / autogenerate Unity Atoms specific script templates


AdamRamberg:
**Is your feature request related to a problem? Please describe.**
This is an issue that tries to increase the developer experience using Unity Atoms. It would be really nice if there where templates for some Atoms, eg: 
- Functions - eg: `Create IntFunction` or `Create BoolIntFunction`
- Actions - eg: `Create IntAction`
- Listeners - eg: `Create FloatBoolListener`
- Events - eg: `Create IntBoolEvent`

My thinking is that these should be generated based on the Atom types that are defined. This should probably also be part of the Generator as well as further additions to the generation (see issue #60). 

**Describe the solution you'd like**
The straight forward solution would be to generate these templates in the Generator following this: https://support.unity3d.com/hc/en-us/articles/210223733-How-to-customize-Unity-script-templates

There should be an option to "Generate Unity Atoms scripting templates", but also "Remove generated Unity Atoms scripting templates". 

**Describe alternatives you've considered**
It would be great if some more investigation were made on if there is an alternative to create templates on a per project basis. Haven't found any solution for this when I've looked around on the Internetz. 

**Additional context**
Consider finishing #60 before starting this issue.

soraphis:
**problems with this**
- ScriptTemplates do not work within packages, so we need to find a way to extract them.
  - we could add it as an exportable "sample" in the package. (https://docs.unity3d.com/Manual/cus-samples.html)
- Each Atom (Function/Action/Listener/Event) would require a lot of templates for different types, I'm not sure if this would be worth it, or if each would have a "generic" template, where the types are to be filled out.

soraphis:
Closing this for now, since no clear development path, missing activity, missing requests and it would be probably superseeded by #495 (e.g. we could create a custom menu, that let's you select a base type and types for each generic, to be filled)

Issue #60, opened 2019-10-16
Generate Atoms of different types


AdamRamberg:
**Is your feature request related to a problem? Please describe.**
Make it possible to generate Atoms of different type for: 
- Actions
- Listeners
- Functions
- Events

Eg. generate a BoolIntEvent. This should be done in the Generator already available in Unity Atoms. 

**Describe the solution you'd like**
This is not part of the regular Atoms generation currently implemented in Unity Atoms. I propose to create a new "page" in the Generator to generate "Multi Atoms" (I'm open for better naming here). The user will first pick what Atom to generate (selection box): Action, Listener, Function or Event. Then options to add types will appear. 

**Describe alternatives you've considered**
Another consideration is to create a new Generator for this kind of generation. 

I would be really appreciate someone making a PR on this. Make sure to read the [guidelines](https://github.com/AdamRamberg/unity-atoms/blob/canary/CONTRIBUTING.md) before starting + assign yourself (or ask me) to the issue. Also don't hesitate to ask any further questions below before developing.

soraphis:
superseeded by #495

PR #281, opened 2021-06-09
Generator rewrite

CaseyHofland:
Boy oh boy is this a doozy.

### The generator has been rewritten to work with types. 
This works brilliantly as you may now perform: 
`Generator.Generate(typeof(AtomEvent<float>), myPath, myNamespace);` 
And things will just work.

### The generator accepts a delegate of type Resolver.
`string Resolver(Type type, out string className, string withNamespace = default)`
The coup de grâce of this is the static Template.ResolveAtom function which generates the atoms entirely using strings (I WOULD use CodeDOM for it but nooooo, that isn't supported by .NET Standard 2.0).
This change means that the templates in the resources folder are no longer needed.

### Resolvers are superuperduper easy to extend.
```
public static string ResolveConstant(Type type, out string className, string withNamespace = default)
{
    type = typeof(AtomBaseVariable<>).MakeGenericType(type);
    return Template.ResolveAtom(type, out className, withNamespace);
}
```
This can then be used like so:
`Generator.Generate(typeof(float), myPath, ConstantTemplate.ResolveConstant, myNamespace);` same as:
`Generator.Generate(typeof(AtomBaseVariable<float>), myPath, myNamespace);`
The gain here is that if you have multiple resolvers, you can pass in the same type for all of them.

### The AtomGenerator options are dynamic and extendable.
Currently the AtomGenerator has options for Constant, Event and Pair Event. These aren't hard coded in: they are retrieved from a static dictionary of resolvers. Say you want to add a new generator. All you have to do is:
```
[InitializeOnLoad]
public static class ConstantTemplate
{
    static ConstantTemplate()
    {
        AtomGenerator.resolvers.Add("Constant", ResolveConstant);
    }

    public static string ResolveConstant(Type type, out string className, string withNamespace = default)
    ...
}
```
This automatically adds the option in the AtomGenerator, thus making extending the AtomGenerator as easy as using it!

In light of these improvements I have deprecated the old generator methods.

This is currently a draft, as it is not yet up to feature-parity with the old generator. For some of the smaller details I would refer you to the changelogs.


Casey-Hofland:
Mentioning #61 since the Resolvers extendibility could probably solve this issue, but it would need to be experimented with.

Casey-Hofland:
Also mentioning #60 for the same reason as above.

Casey-Hofland:
Mentioning #235 because this update should totally make the AtomGenerator good for it.

Casey-Hofland:
Creating the templates for the new system is underway, but I will need to test if they are created exactly like the old ones.

Casey-Hofland:
I have a question regarding the naming of atoms. Currently I generate as much as I can just from the Type you are generating from. For something like Constants, this leads to class names like Bool2AtomBaseVariable, since what we call a Constant is actually the class AtomBaseVariable.
I do prefer this method since I think it makes it clear where a generated atom originates from. However, it would be better I think if we rename types like AtomBaseVariable to simply Constant, which will generate the desired name Bool2Constant. Remember, it's still in the UnityAtoms namespace, so its full name is still UnityAtoms.Constant.

miikalo:
> I have a question regarding the naming of atoms. Currently I generate as much as I can just from the Type you are generating from. For something like Constants, this leads to class names like Bool2AtomBaseVariable, since what we call a Constant is actually the class AtomBaseVariable.
> I do prefer this method since I think it makes it clear where a generated atom originates from and incentivizes using good naming conventions. Then, it would be better I think if we rename types like AtomBaseVariable to simply Constant, which will generate the desired name Bool2Constant. Remember, it's still in the UnityAtoms namespace, so its full name is still UnityAtoms.Constant. Would this solution be acceptable?

I like the idea of simplifying names, but even though Constant does not add any functionality to an AtomBaseVariable, either of them might change later for different reasons, and that's why I wouldn't mind being a bit more explicit. How about UnityAtoms.BaseVariable?

miikalo:
Nitpicking, but:

image

This was kind of confusing, not sure why it says "Mono Script" and I wasn't sure what to select for it, until I clicked Generate and the generated Atom appeared in it's place. When you first see the generator it's confusing. Could it be a read-only field but it would still take you to the generated asset when clicked?

Casey-Hofland:
> Could it be a read-only field but it would still take you to the generated asset when clicked?

Sure thing! I’ve experienced some confusion from it myself and this is actually the perfect solution.

Casey-Hofland:
I’ll come back to the rest later when I’m behind my laptop. Don’t comment or quote this message.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -21,6 +22,191 @@ public static string[] GetTemplatePaths()
...
+            if (asPair)
+            {
+                typeName = $"Pair<{typeName}>";
+                displayName += "Pair";
+            }
+
+            // Add Types.
+            var templateVariables = new Dictionary<string, string>()
+            {
+                { "VALUE_TYPE", typeName },
```
miikalo, line 40: Hard-coded string to enums, would it make sense?

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -21,6 +22,191 @@ public static string[] GetTemplatePaths()
...
+            // Add type.
+            var templateConditions = new List<string>
+            {
+                $"TYPE_IS_{typeName.ToUpper()}"
+            };
+
+            // Add Type code condition if applicable.
+            if (type.IsNumeric())
+            {
+                templateConditions.Add("IS_NUMERIC");
```
miikalo, line 72: Same here about strins to enums.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -21,6 +22,191 @@ public static string[] GetTemplatePaths()
...
+            }
+            if (!string.IsNullOrEmpty(withNamespace))
+            {
+                templateVariables.Add("SUB_UA_NAMESPACE", withNamespace);
+            }
+
+            return templateVariables;
+        }
+
+        public static List<string> CreateTemplateConditions(Type type, string withNamespace = default)
```
miikalo, line 59: Missing documentation for the method, the Generate method has good documentation, as an example.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -21,6 +22,191 @@ public static string[] GetTemplatePaths()
             return Directory.GetFiles(templateSearchPath, "UA_Template*.txt", SearchOption.AllDirectories);
         }
 
+        public static Dictionary<string, string> CreateTemplateVariablesMap(Type type, bool asPair, string withNamespace = default)
```
miikalo, line 25: Documentation would go a long way to explain the intent of the method.

Packages/Core/Editor/Generator/Templates/Template.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            public string namespaces;
+            public string withNamespace;
+            public string typeName;
+            public string attributes;
+            public string nestedMenu;
+        }
+
+        public const string autoGeneratedWarning =
+@"//------------------------------------------------------------------------------
```
miikalo: We can work on this comment field to make it clearer that it's by Unity Atoms, generated in this version, etc etc, but it's not hugely important right now. Maybe drop a todo here?

miikalo (approved):
Great job, the bulk of work seems to be out of the way with this!

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -21,6 +21,9 @@ public static string[] GetTemplatePaths()
             return Directory.GetFiles(templateSearchPath, "UA_Template*.txt", SearchOption.AllDirectories);
         }
 
+
+        #region Deprecated
```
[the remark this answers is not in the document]
Casey-Hofland: I wasn’t sure, but I thought I’d at least keep it in until all other atoms packages have been converted as well

Packages/Core/Editor/Extensions/TypeExtensions.cs
```diff
@@ -1,4 +1,5 @@
 using System;
```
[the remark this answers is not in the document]
Casey-Hofland: Haha, it’s more extensive on my local git, but as I’m still working on the other templates I haven’t pushed those changes yet.

Packages/Core/Editor/Generator/AtomType.cs
```diff
@@ -31,18 +31,20 @@ internal struct AtomType : IEquatable<AtomType>
             string drawerTemplateName = "",
             string relativeDrawerFileNameAndPath = "",
             string editorTemplateName = "",
-            string relativeEditorFileNameAndPath = ""
+            string relativeEditorFileNameAndPath = "",
+            bool isValuePair = false
```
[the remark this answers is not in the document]
Casey-Hofland: For if the type needs to be generated as a Pair (like EventPair). It didn’t really last long though as I quickly deprecated this script in place of the Resolver system I have in place now.

Packages/Core/Editor/Generator/AtomRecipe.cs
```diff
@@ -5,8 +5,8 @@ namespace UnityAtoms.Editor
     /// <summary>
     /// Internal module class that holds that regarding an Atom type.
     /// </summary>
-
-    internal struct AtomReceipe : IEquatable<AtomReceipe>
+    [Obsolete("Atom Receipe is outdated.", false)]
```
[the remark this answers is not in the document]
Casey-Hofland: Haha I didn’t even notice, that one is pretty easy though so why not 😅

Packages/Core/Editor/Generator/Generator.cs
Casey-Hofland: These methods actually allow you to use the old generator with types, which was my first approach. These will also be deprecated (which is funny cause they’re new) before I make this pr ready for review.

Packages/Core/Editor/Generator/Generator.cs
Casey-Hofland: Same as above.

Packages/Core/Editor/Generator/Generator.cs
Casey-Hofland: Same as above.

Packages/Core/Editor/Generator/Generator.cs
Casey-Hofland: Samesies

Packages/Core/Editor/Generator/Templates/Template.cs
Casey-Hofland: Good idea, I’ll drop the todo and see if I can do this dynamically because I wouldn’t want to have to change it with every new version.

Casey-Hofland: Putting this here so I don't forget it: [PackageInfo.FindForAssetPath](https://docs.unity3d.com/2020.1/Documentation/ScriptReference/PackageManager.PackageInfo.FindForAssetPath.html)(AssetDatabase.GetAssetPath(this));
This _may_ work and give us all kinds of package information to include.

Casey-Hofland:
Quick update on plans:

This generator merge will likely not finish it yet. I've taken another good look at my goal, and what is very important to me is that the generator needs to be able to generate all the atoms in base atoms on its own.

This creates a bit of a dilemma because that is currently not 'entirely' possible. Base atoms still needs a few updates here and there for the V5 change, which will thus influence the generator as well.

Plus, now that the generator is more general, I want to experiment with creating e.g. a ClampGenerator, but this is no priority of mine and I might leave it for after 5.0.0.

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -257,7 +236,7 @@ public static List<string> CreateTemplateConditions(bool isValueTypeEquatable, s
         /// </code>
         /// </example>
         [Obsolete("Use the same method with a Type parameter instead.", false)]
-        public static void Generate(AtomReceipe atomReceipe, string baseWritePath, string[] templatePaths, List<string> templateConditions, Dictionary<string, string> templateVariables)
+        public static void Generate(AtomRecipe atomReceipe, string baseWritePath, string[] templatePaths, List<string> templateConditions, Dictionary<string, string> templateVariables)
         {
             var (atomType, valueType) = atomReceipe;
```
miikalo, line 241: Receipe still lurks on this line... 😄

miikalo:
Good work so far, just request a re-review when ready.

Packages/Core/Editor/Generator/Generator.cs
Casey-Hofland: Nice catch, I've fixed it locally so it will come around one of these updates.

Packages/Core/Editor/Generator/Templates/ConstantTemplateFactory.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Collections.Generic;
+using UnityEditor;
+
+namespace UnityAtoms.Editor
+{
+    [InitializeOnLoad]
+    public static class ConstantTemplateFactory
```
miikalo, line 8: All these factories seem to share the same methods, but they don't implement an interface or force the methods in any other way either. Any solutions for this in C#? Do these need to be static?

Packages/Core/Editor/Generator/AtomGenerator.cs
```diff
@@ -1,86 +1,75 @@
...
-
-        private void Reset()
+        private static readonly object populatorsLock = new object();
+        public static event Populator populators
         {
-            for(int i = 0; i < AtomTypes.ALL_ATOM_TYPES.Count; i++)
+            add
             {
-                GenerationOptions |= (1 << i);
+                var generatorGuids = AssetDatabase.FindAssets($"t:{nameof(AtomGenerator)}");
```
miikalo, line 15: This part repeats, as well as the loop below it. Extract to private methods or move to base class.

Packages/Core/Editor/Generator/Templates/PairTemplateFactory.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using UnityEngine;
+
+namespace UnityAtoms.Editor
+{
+    public static class PairTemplateFactory
```
miikalo, line 6: Why does this one look so different from the others?

Packages/Core/Editor/Generator/Templates/VariableTemplateFactory.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                solver.template.baseClass = typeof(EquatableAtomVariable<>).MakeGenericType(generatedType);
+
+                var body = string.Empty;
+
+                if (generatedType.IsColor())
+                {
+                    body =
+$@"
+        /// <summary>
```
miikalo, line 39: What are these methods? I don't remember them...

Packages/Core/Editor/Generator/ClassModifier.cs
```diff
@@ -0,0 +1,9 @@
+namespace UnityAtoms.Editor
+{
+    public enum ClassModifier
```
miikalo, line 3: Virtuals, Partials, how many need to be here?

miikalo (commented):
Looks good, really like the ideas presented in the video. Not going to approve before the PR goes live.

Packages/Core/Editor/Generator/Templates/VariableTemplateFactory.cs
```diff
@@ -0,0 +1,164 @@
...
+                }
+                else if (generatedType.IsNumeric())
+                {
+                    body =
+$@"
+        /// <summary>
+        /// Add value to Variable.
+        /// </summary>
+        /// <param name=""value"">Value to add.</param>
+        public void Add({generatedType.GenericName()} value) => Value += value;
```
AdamRamberg, line 60: I believe we should use extension methods of the generic type instead of defining these in the generator. Eg `public static void Add(this AtomVariable<int> variable, int value) => variable.Value += value;`. That way we can keep the extension in the UnityAtoms core package and don't need to regenerate it every time and we avoid keeping the implementation as a string in this template.

AdamRamberg: Disregard my comment above. Extension methods are not accessible from UnityEvents so it is not a good solution.

Packages/Core/Editor/Generator/ClassModifier.cs
Casey-Hofland: Partial would be a different kind of modifier, since a class can be "partial sealed" or "partial abstract".

Packages/Core/Editor/Generator/Templates/VariableTemplateFactory.cs
Casey-Hofland: They were in the original .txt templates. Under certain conditions, the generator will create methods for the generated AtomVariable as well.

Packages/Core/Editor/Generator/Templates/PairTemplateFactory.cs
Casey-Hofland: Instead of writing a GetPairTemplate method for every pair, I instead wrote a GetTemplate method that modifies an existing GetTemplate method to turn that template into a pair. That may be pretty confusing, but just look how it is used in the EventTemplateFactory. It generates both an AtomEvent and an AtomEventPair from the same GetTemplate method, but one is parsed through the PairTemplateFactory.

Packages/Core/Editor/Generator/Templates/ConstantTemplateFactory.cs
Casey-Hofland: The way I implemented it, the GetTemplate methods need to be static in order to be added dynamically to the AtomGenerator (in hindsight, perhaps ScriptableObjects carrying Templates would have been better).
Since everything in the Factories was static and I didn't want them to be instantiatable, I opted to make them static.

Packages/Core/Editor/Generator/Templates/VariableTemplateFactory.cs
Casey-Hofland: Bummer, it was a nice dream.

Packages/Core/Editor/Generator/ClassModifier.cs
Casey-Hofland: Keep in mind that "public, private, protected" and similar would be an access modifier. I didn't write this whole generator system out as much since we don't need all that for Atoms and since actually writing out such a system would take a lot of work getting the details right.

Packages/Core/Editor/Generator/AtomGenerator.cs
Casey-Hofland: Good point, I will make it a private method (the Generator needs to stay clean of this since the AtomGenerator uses this as a kind of "convenience event": no need for the Factories to repeat this FindAssets method every time they want to add their Template.

Casey-Hofland:
The summary option has already been tested and generates the exact same summaries as the old generator.

Packages/Core/Editor/Generator/Templates/AtomTemplateFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+            template.attributes.Add(typeof(CreateAssetMenuAttribute), new[] { $"fileName = \"{template.className}\"", $"menuName = \"Atoms/{template.className.Remove(atomName.Remove(" "))}/{atomName}\"" });
+
+            return template;
+        }
+
+        public static Template GetComponentTemplate(Type atomType, Type typeArgument, string atomName, string editorIconName)
+        {
+            var template = GetTemplate(atomType, typeArgument, atomName, editorIconName);
+
+            template.attributes.Add(typeof(AddComponentMenu), new[] { $"\"Atoms/{template.className.Remove(atomName.Remove(" "))}/{atomName}\"" });
```
AdamRamberg, line 37: the last `/atomName` is not enough here since this is going to be the display name of components added to GameObjects (tests it yourself). I would suggest the last part to be `template.className`, but split on camel case (something like https://stackoverflow.com/a/6137889).

Packages/Core/Editor/Generator/Templates/AtomTemplateFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            return template;
+        }
+
+        public static Template GetAssetTemplate(Type atomType, Type typeArgument, string atomName, string editorIconName)
+        {
+            var template = GetTemplate(atomType, typeArgument, atomName, editorIconName);
+
+            template.classModifier = ClassModifier.Sealed;
+            template.attributes.Add(typeof(CreateAssetMenuAttribute), new[] { $"fileName = \"{template.className}\"", $"menuName = \"Atoms/{template.className.Remove(atomName.Remove(" "))}/{atomName}\"" });
```
AdamRamberg, line 28: Make sure that the "namespace" "Atoms" is used everywhere. Currently we are using the "namespace" "Unity Atoms" for all Atoms.

Packages/Core/Editor/Generator/Templates/PairTemplateFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public static Template GetTemplate(Type type, Func<Type, Template> templateFactory)
+        {
+            var pairType = typeof(Pair<>).MakeGenericType(type);
+
+            var template = templateFactory.Invoke(pairType);
+            var typeName = type.CSharpName().Capitalize();
+            var pairTypeName = pairType.CSharpName().Capitalize();
+
+            template.className = template.className.Replace(pairTypeName, typeName) + atomName;
```
AdamRamberg, line 19: Appending Pair to the end does not seem right to me. The result becomes for example `ByteEventInstancerPair`, but the desired result is `BytePairEventInstancer`.

Packages/Core/Editor/Generator/Templates/PairTemplateFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+            template.summary = template.summary.Remove(template.summary.IndexOf(' '));
+            template.summary += $" of type {WebUtility.HtmlEncode(pairType.CodeCompatibleName())}. Inherits from {WebUtility.HtmlEncode(template.baseClass.CodeCompatibleName())}.";
+
+            if (template.attributes.ContainsKey(typeof(CreateAssetMenuAttribute)))
+            {
+                var fileName = template.attributes[typeof(CreateAssetMenuAttribute)][0];
+                fileName = $"{fileName.Replace(pairTypeName, typeName)} + \"{atomName}\"";
+                template.attributes[typeof(CreateAssetMenuAttribute)][0] = fileName;
```
AdamRamberg, line 27: Same problem as `className` above.

Packages/Core/Editor/Generator/Templates/PairTemplateFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+
+                var menuName = template.attributes[typeof(CreateAssetMenuAttribute)][1];
+                menuName = $"{menuName.Replace(pairTypeName, typeName)} + \" {atomName}\"";
+                template.attributes[typeof(CreateAssetMenuAttribute)][1] = menuName;
```
AdamRamberg, line 31: Same problem as `className` above.

Packages/Core/Editor/Generator/Templates/PairTemplateFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+            if (template.attributes.ContainsKey(typeof(AddComponentMenu)))
+            {
+                var menuName = template.attributes[typeof(AddComponentMenu)][0];
+                menuName = $"{menuName.Replace(pairTypeName, typeName)} + \" {atomName}\"";
+                template.attributes[typeof(AddComponentMenu)][0] = menuName;
```
AdamRamberg, line 38: Same problem as `className` above.

AdamRamberg (changes requested):
Really great job on this PR! See my comments for some stuff that needs to be addressed before we merge. Other than the comments we also need to create generators for all generated types in all sub packages (BaseAtoms etc). This was handled by the old generator and needs to be in place before we merge (or at least before release of v5).

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -1,164 +1,232 @@
...
         /// </summary>
-        /// <param name="template">The content template to remove namespace from.</param>
-        /// <returns>A copy of `content`, but without duplicate namespaces.</returns>
-        private static string RemoveDuplicateNamespaces(string template)
+        public void Generate()
         {
-            var currentIndex = 0;
-            var countNamespaces = new Dictionary<string, int>();
-            while (true)
+            var assetPaths = new string[solvers.Values.Count];
```
AdamRamberg, line 137: If we generate a set of Atoms of one type, and then change the type and regenerate the old Atoms are still there, but unreferenced by the the generator and the generator creates a new set of Atoms. In my mind we should delete the old Atoms of the old type and then create the new ones. Maybe we should show a prompt before regenerating as well, where one could choose to delete or to keep the Atoms (unreferenced by any generator)?

Packages/Core/Editor/Generator/Generator.cs
```diff
@@ -1,164 +1,232 @@
...
             }
+        }
+
+        /// <summary>
+        /// Generate scripts inside the directory from the input templates.
+        /// </summary>
+        /// <param name="directory">The folder path where the scripts are going to be written to.</param>
+        /// <param name="template">The template to generate the script from.</param>
+        /// <param name="templates">The templates to generate the scripts from.</param>
+        public static MonoScript[] Generate(string directory, params Template[] templates)
```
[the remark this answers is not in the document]
Casey-Hofland: I would like to keep it in there as a general method. The idea behind it is that if you have a Generate(string, Template) method somewhere and ever want to extend it, you can just do Generate(string, Template, Template, Template) and it will work.

Packages/Core/Editor/Editors/GeneratorEditor.cs
```diff
@@ -0,0 +1,209 @@
...
+                var script = _scripts.GetArrayElementAtIndex(i);
+
+                var optionLabel = new GUIContent(key.stringValue);
+
+                EditorGUILayout.BeginHorizontal();
+
+                EditorGUILayout.PropertyField(option, optionLabel);
+
+                EditorGUI.BeginDisabledGroup(true);
+                //EditorGUILayout.PropertyField(script, GUIContent.none);
```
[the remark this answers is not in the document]
Casey-Hofland: Ah no, should be removed. I still need to take out the broken MonoScripts from the Generator.

Packages/Core/Editor/Generator/Generator.cs
Casey-Hofland: Unfortunately, since the MonoScript references will have to be removed because they are broken, I can't implement this feature.

Packages/Core/Editor/Generator/Generator.cs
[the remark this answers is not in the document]
AdamRamberg: Alright, that is fine

Packages/Core/Editor/Editors/GeneratorEditor.cs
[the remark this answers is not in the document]
AdamRamberg: Why are MonoScripts broken? Not sure if I've missed a previous discussion?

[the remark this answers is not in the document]
Casey-Hofland: I do some weird stuff in OnBefore- and OnAfterSerialization which breaks it unfortunately.

PR #289, opened 2021-07-09
Fix hasChildren display in AtomReferences

Casey-Hofland:
The display of an AtomReference would act a bit weird if the Atom was of e.g. a serializable class.

Old: open button is displayed weirdly on top of the selection button.
Screenshot 2021-07-09 141001

New: open button is displayed at the original position. 
Screenshot 2021-07-09 141028

Since the EditorGUI.PropertyField in question has GUIContent.none, I do not expect this to break any other displays.

### DIFF
--- a/Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs
+++ b/Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs
@@ -67,16 +67,12 @@ namespace UnityAtoms.Editor
             var newUsageValue = EditorGUI.Popup(buttonRect, currentUsage.intValue, GetPopupOptions(property), _popupStyle);
             currentUsage.intValue = newUsageValue;
 
-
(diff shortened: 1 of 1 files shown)


miikalo:
Can't seem to be able to replicate this, could you be more specific about reproduction steps?

Casey-Hofland:
Create an atomreference of a serializable class.

e.g.
`
[Serializable]
public class Foo {}

(Then generate:)
FooAtomReference : AtomReference<Foo>
`

If you then use FooAtomReference, it would look weird in the inspector before this fix.

miikalo (approved):

miikalo:
> Create an atomreference of a serializable class.
> 
> e.g.
> 
> ```
> [Serializable]
> public class Foo {}
> 
> (Then generate)
> FooAtomReference : AtomReference<Foo> {}
> ```
> 
> If you then use FooAtomReference, it would look weird in the inspector before this fix.

Got it reproduced by adding a public int abc in Foo. The PR works as intended, I'm merging it.

PR #297, opened 2021-09-14
Fix #296 FiniteStateMachine not listening to Raise Event To Complete Transitio…

### LINKED ISSUE #296: [BUG] FiniteStateMachine not listening to Raise Event To Complete Transition
When using FSM with Raise Event To Complete Transition on, and doing a manual raise ( triggered by TransitionStart.Raise), the Complete Raise is missed.

After reviewing the FSM code, I was able to fix it by calling the method that registers the listener BEFORE the Raise of Transition start at [FiniteStateMachine.cs#285](https://github.com/unity-atoms/unity-atoms/blob/7b6b8431703b557647cd11e6632e0df8888fb8db/Packages/FSM/Runtime/FiniteStateMachine/FiniteStateMachine.cs#L285):
Before the fix:
`
 if (transition.TestCondition())
                {
                    _currentTransition = transition;
                    if (_transitionStarted != null)
                    {
                        _transitionStarted.Raise(
                            new FSMTransitionData()
                            {
                                FromState = transition.FromState,
                                ToState = transition.ToState,
                                Command = transition.Command,
                                CompleteTransition = _completeCurrentTransition,
                            }
                        );
                    }
                    **transition.Begin(this, EndCurrentTransition);**
                }
`

After the fix:
`
 if (transition.TestCondition())
                {
                    _currentTransition = transition;
                    transition.Begin(this, EndCurrentTransition);
                    if (_transitionStarted != null)
                    {
                        _transitionStarted.Raise(
                            new FSMTransitionData()
                            {
                                FromState = transition.FromState,
                                ToState = transition.ToState,
                                Command = transition.Command,
                                CompleteTransition = _completeCurrentTransition,
                            }
                        );
                    }
                }
`

Basicaly transition.Begin calls `_fsmReference.CompleteCurrentTransition.RegisterListener(this);` and  _transitionStarted.Raise will Raise this event. So if the register happens after the raise, it wont be listening.

Would be also great to have on the wiki stating that the Raise must be called with a bool parameter for true or false when doing it manually. Since its possible to call a Raise without parameters, that (as far as I checked) wont work, and it's not clear if we need to raise the event only when we accpeted it, or if we should when it fails to change state (what is the case)

Kyrw:
When using FSM with Raise Event To Complete Transition on, and doing a manual raise ( triggered by TransitionStart.Raise), the Complete Raise is missed.

After reviewing the FSM code, I was able to fix it by calling the method that registers the listener BEFORE the Raise of Transition start at FiniteStateMachine.cs#285:
Before the fix:
if (transition.TestCondition()) { _currentTransition = transition; if (_transitionStarted != null) { _transitionStarted.Raise( new FSMTransitionData() { FromState = transition.FromState, ToState = transition.ToState, Command = transition.Command, CompleteTransition = _completeCurrentTransition, } ); } **transition.Begin(this, EndCurrentTransition);** }

After the fix:
if (transition.TestCondition()) { _currentTransition = transition; transition.Begin(this, EndCurrentTransition); if (_transitionStarted != null) { _transitionStarted.Raise( new FSMTransitionData() { FromState = transition.FromState, ToState = transition.ToState, Command = transition.Command, CompleteTransition = _completeCurrentTransition, } ); } }

Basicaly transition.Begin calls _fsmReference.CompleteCurrentTransition.RegisterListener(this); and _transitionStarted.Raise will Raise this event. So if the register happens after the raise, it wont be listening.

### DIFF
--- a/Packages/FSM/Runtime/FiniteStateMachine/FiniteStateMachine.cs
+++ b/Packages/FSM/Runtime/FiniteStateMachine/FiniteStateMachine.cs
@@ -270,6 +270,7 @@ namespace UnityAtoms.FSM
                 if (transition.TestCondition())
                 {
                     _currentTransition = transition;
+                    transition.Begin(this, EndCurrentTransition);
                     if (_transitionStarted != null)
                     {
                         _transitionStarted.Raise(
@@ -282,7 +283,6 @@ namespace UnityAtoms.FSM
                             }
                         );
                     }
-                    transition.Begin(this, EndCurrentTransition);
                 }
             }
             else


miikalo (approved):

miikalo:
Seems to work as intended, I changed the base branch to canary and brought in the description from the bug report.

PR #301, opened 2021-09-27
Handling a denied transition.

Kyrw:
Consider as if the transtion request never happened


miikalo:
Thanks for the suggestion, I'll put this in v5 because it changes the behavior of Atoms.

soraphis:

PR #307, opened 2021-10-24
#243 Add scope (or lifetime) management (updated for V5)

bguyl:
Updated version of the following PR: https://github.com/unity-atoms/unity-atoms/pull/248

### DIFF
--- a/Packages/Core/Editor/Editors/Variables/AtomVariableEditor.cs
+++ b/Packages/Core/Editor/Editors/Variables/AtomVariableEditor.cs
@@ -26,6 +26,10 @@ namespace UnityAtoms.Editor
         {
             serializedObject.Update();
 
+            DisplayScopeDropdown();
+
+            GUI.enabled = true;
+
             EditorGUILayout.PropertyField(serializedObject.FindProperty("_developerDescription"));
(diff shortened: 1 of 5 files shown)


toasterhead-master:
I know I'm late, but I'm reading it just now. Maybe the solution for that problem is to just include a hint underneath the selector for developers/designers to understand what each option does. That way we keep the simplistic naming and avoid confusion.

soraphis:
hm... imho the best solution to this would be to remove the runtime state from the scriptable object alltogether.
save the state instead in a global-singleton in a dictionary. 

since the global-singleton is DontDestroyOnLoad we don't have the issue anymore that switching between scenes kills the state in a build but not in the editor (or vice versa).
we also don't have the problem anymore that using asset bundles / addressables might load multiple instances of the same asset, which all have their unique value. (which this solution does not address at all, IIRC)

we also have the oportunity to create a "scope"-field on variables/instancers.
- if empty state is stored on the singleton
- if pointing to a gameobject, state is stored on that gameobject (allows for level-scoped variables)

without having tested in thoroughly, this is currently my prefered solution for this issue

soraphis:

PR #309, opened 2021-11-22
Fix #255 AtomVariableInstancer assumes Base != null

### LINKED ISSUE #255: [BUG] AtomVariableInstancer assumes Base != null
**Describe the bug**
AtomVariableInstancer assumes Base != null. Simple issue just add any VariableInstancer, for instance ColorVariableInstancer on a GameObject. When you enter Play mode an error will trigger in the ImplSpecificSetup. 

**To Reproduce**
1. Create a new scene.
2. In the new scene create a new GameObject.
3. Add a ColorVariableInstancer component to the new GameObject.
4. Enter Play mode.
5. An error will trigger.

**Expected behavior**
I would expect the error to not trigger. The AtomBaseVariableInstancer class seems capable of knowing how to handle the case when Base is null, so it seems odd that the AtomVariableInstancer cannot also handle that. It seems that a null check before the Base.Changed would be enough to alleviate the issue.

**Environment**
 - OS: Using Mac Catalina, but this is a very self evident error that will trigger on all platforms.
 - Unity Version: 2020.2.5f1
 - Version of Unity Atoms: 4.4.2

ThimoDEV:
The null check on Base events in AtomVariableInstancer didn't account for Base itself and didnt create runtime AtomEvents.

### DIFF
--- a/Packages/Core/Runtime/VariableInstancers/AtomVariableInstancer.cs
+++ b/Packages/Core/Runtime/VariableInstancers/AtomVariableInstancer.cs
@@ -29,22 +29,14 @@ namespace UnityAtoms
         /// </summary>
         protected override void ImplSpecificSetup()
         {
-            if(Base != null)
+            if (Base.Changed != null)
             {
-                if (Base.Changed != null)
-                {
-                    _inMemoryCopy.Changed = Instantiate(Base.Changed);
-                }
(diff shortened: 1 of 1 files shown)


miikalo:
Squashed because of revert commits.

PR #311, opened 2021-11-22
fix #254 [BUG] ulong variable throws error when the value being edited in play mode.

### LINKED ISSUE #254: [BUG] `ulong` variable throws error when the value being edited in play mode.
**Describe the bug**
I generated atoms with `ulong` as the type. Then I made an `ulong` variable. Everytime I change the value on play mode, I got this error:

```
InvalidCastException: Specified cast is not valid.
UnityAtoms.AtomBaseVariable`1[T].set_BaseValue (System.Object value) (at Library/PackageCache/com.unity-atoms.unity-atoms-core@4.4.2/Runtime/Variables/AtomBaseVariable.cs:48)
UnityAtoms.Editor.AtomVariableEditor`2[T,P].OnInspectorGUI () (at Library/PackageCache/com.unity-atoms.unity-atoms-core@4.4.2/Editor/Editors/Variables/AtomVariableEditor.cs:53)
UnityEditor.UIElements.InspectorElement+<>c__DisplayClass58_0.<CreateIMGUIInspectorFromEditor>b__0 () (at <e020ad89c1af43769f98694c4076cf19>:0)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr, Boolean&)
```

**To Reproduce**
1. Generate atoms for type `ulong`.
2. Make new `ulong` variable.
3. Start play mode.
4. Change value of the variable.

**Expected behavior**
No error and event raised.

**Environment**
 - OS: Windows 10
 - Unity Version 2020.2.1f1
 - Version of Unity Atoms v4.4.2

ThimoDEV:
Ulong was deserialized to a System32 int. The solution was to use the same fix as for the Double type which was serialized to a float. I guess more of these issues might appear in the future so then this fix might become quite a long bit of code

### DIFF
--- a/Packages/Core/Editor/Editors/Variables/AtomVariableEditor.cs
+++ b/Packages/Core/Editor/Editors/Variables/AtomVariableEditor.cs
@@ -48,6 +48,11 @@ namespace UnityAtoms.Editor
                         {
                             atomTarget.BaseValue = (double)(float)value;
                         }
+                        //Ulong is deserialized to System32 Int. 
+                        else if(typeof(T) == typeof(ulong))
+                        {
+                            atomTarget.BaseValue = (ulong)(int)value;
+                        }
                         else
                         {
                             atomTarget.BaseValue = value;


miikalo (approved):
This removes the error, but I wonder if the ulong value gets truncated when cast to int. This seems a problem in Unity serialization rather than something we can fix.

PR #313, opened 2021-11-23
fix #282 changed/changedwithhistory events not created when registere…

### LINKED ISSUE #282: [BUG] Default Changed and ChangedWithHistory Events Not Created at Runtime
Hi all, love the framework and all the hard work everyone has done here!

The PR from #179 is a great addition and makes using the library easier.  Since 4.4.2 it looks like these default values are no longer created at runtime when Registering within a script. Runtime events *are* created if they are added to a Reference Listener component.

**To Reproduce**
1. Create a new Variable, any type will do (e.g. a BoolVariable)
2. Create a new Monobehavior (seen below) and place the variable in the property drawer.
3. Start Play mode
4. Check the console or check the BoolVariable's property window.

**Expected behavior**
The `Changed` event should not be null, and should be filled with `Bool Variable_Changed Event_Runtime_Unity Atoms.Base Atoms.Bool Event`

**Screenshots**
Unity Atoms 4.3.0 (runtime event created)
UnityAtoms-4 3 0

Unity Atoms 4.3.0 (runtime event missing)
UnityAtoms-4 4 3

**Environment**
 - OS: Windows 10
 - Unity Version: tested with 2019.4 and 2020.3
 - Version of Unity Atoms: 4.4.3

**Additional context**
TestBehavior.cs

```
using UnityEngine;
using UnityAtoms.BaseAtoms;

public class TestBehavior : MonoBehaviour
{
    [SerializeField] private BoolVariable Bool;

    private void Awake() {
        if (Bool != null && Bool.Changed == null)
            print("Bool.Changed is null");
    }
}
```

ThimoDEV:
The problem was indeed that the changed and changedWithHistory events where not created when called in code. There are 3 possible solutions I can come up with to fix this bug. One is to slightly modify the GetOrCreate<E>() method so it can be called in the property. 

Option two is to call the runtime creation code in the property itself again. This makes the GetOrCreate<E>() method unnessesary I think?

Option 3 is to somehow add an usage for calling a variables events in code

Maybe another solution will work better than what I suggested here in this PR. Im all open for the best suggestion to tackle this problem and to update this PR.

### DIFF
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -52,6 +52,7 @@ namespace UnityAtoms
         {
             get
             {
+                GetOrCreateEvent<E1>();
                 return _changed;
             }
             set
@@ -71,6 +72,7 @@ namespace UnityAtoms
         {
             get
             {
+                GetOrCreateEvent<E2>();
                 return _changedWithHistory;
             }
             set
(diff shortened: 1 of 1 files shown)


miikalo (approved):
Tested on 2019.4.21f1, works as described.

boynukaline:
Hey this overrides the current change event when a change event is added to a variable. Steps to reproduce:
1. Leave variable changed event blank and in runtime change variable value to something. Stop runtime.
2. Add your change event to the variable, and go into runtime again and change variable value. Check variable event, it changes to default.

soraphis:
@boynukaline i think it would be better to create a new issue (or reopen the other issue) since this PR is already released 🙈 
thanks for being observant!

boynukaline:
@soraphis Thanks, created new issue!

PR #314, opened 2021-11-23
fix #237 Drawer tries to create field for public Action

### LINKED ISSUE #237: [BUG] Drawer tries to create field for public Action
The problem is that Action is not serializable, but when it's a public variable the drawer tries to create a field for it - which fails

atomerror

Error log message:
ArgumentNullException: Value cannot be null.
Parameter name: property
UnityAtoms.Editor.SerializedPropertyExtensions.GetPropertyValue (UnityEditor.SerializedProperty property) (at Library/PackageCache/com.mambojambostudios.unity-atoms-core@4.4.2/Editor/Extensions/SerializedPropertyExtensions.cs:73)
UnityAtoms.Editor.SerializedPropertyExtensions.GetGenericPropertyValue[T] (UnityEditor.SerializedProperty property, T managedObjectOut) (at Library/PackageCache/com.mambojambostudios.unity-atoms-core@4.4.2/Editor/Extensions/SerializedPropertyExtensions.cs:27)
UnityAtoms.Editor.AtomVariableEditor`2[T,P].OnInspectorGUI () (at Library/PackageCache/com.mambojambostudios.unity-atoms-core@4.4.2/Editor/Editors/Variables/AtomVariableEditor.cs:60)
UnityEditor.UIElements.InspectorElement+<>c__DisplayClass58_0.<CreateIMGUIInspectorFromEditor>b__0 () (at <e020ad89c1af43769f98694c4076cf19>:0)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr, Boolean&)

The class:
https://pastebin.com/XAbr3SMc

ThimoDEV:
The value property inside an AtomVariable is used with the Unity JsonUtility to make sure changes are working properly. There was no check for the System.NonSerialized attribute. This made it igonre that attribute only on the AtomVariable value property. By adding this check the System.NonSerialized attrbute also gets respected in the value. It was already working with this attribute on the C# action in the initialValue.

### DIFF
--- a/Packages/Core/Editor/Extensions/SerializedPropertyExtensions.cs
+++ b/Packages/Core/Editor/Extensions/SerializedPropertyExtensions.cs
@@ -24,7 +24,10 @@ namespace UnityAtoms.Editor
             {
                 try
                 {
-                    field.SetValue(box, property.FindPropertyRelative(field.Name).GetPropertyValue());
+                    if (!field.Attributes.HasFlag(FieldAttributes.NotSerialized))
+                    {
+                        field.SetValue(box, property.FindPropertyRelative(field.Name).GetPropertyValue());
+                    }
                 }
                 catch (InvalidOperationException)
                 {


miikalo (changes requested):
In my testing (Unity 2019.4.21f1) the changes unfortunately do not seem to fix the issue. I have a simple class I generated Atoms for and I tested this by creating a variable and collapsing and uncollapsing the "Value" on the Inspector. I used this class for testing:

```cs
using System;

[Serializable]
public class Bug237
{
    public int number;
    public Action OnValueChanged;
}
```

If I use the class as-is, I get the same error as in the bug. If I comment out the Action variable, I get "The method or operation is not implemented":

image

This was tested with the PR changes.

ThimoDEV:
My fix was to explicitly use the [System.NonSerialized] Tag on the  public Action OnValueChanged;. Its the check that gets done on this value thats causing this problem and I've added this attribute check to avoid the system still trying to serialize the action. Would you mind trying it with this attribute?

miikalo:
Sorry, did not catch that. I'll try again.

miikalo:
> My fix was to explicitly use the [System.NonSerialized] Tag on the public Action OnValueChanged;. Its the check that gets done on this value thats causing this problem and I've added this attribute check to avoid the system still trying to serialize the action. Would you mind trying it with this attribute?

Sorry, did not catch that. I'll try again.

miikalo:
Tried it again, the current change does get rid of the "Value cannot be null" but now I get "The method or operation is not implemented" for the public int. It's probably a different issue that crept in at some point.

miikalo:
I had generated the Atoms with the Equatable flag on by accident. The fix works as intended.

miikalo (approved):

Issue #149, opened 2020-05-04
[FEATURE] `Create` button in variables create events as nested scriptable objects.


mnarimani:
**Is your feature request related to a problem? Please describe.**
You can have multiple scriptable objects in the same asset file. 
Instead of having separate event scriptable objects, we can have events of every variable nested in that variable.
![image](https://user-images.githubusercontent.com/30625116/80993667-b5fce880-8e50-11ea-9997-fa04bc177fc2.png)
![image](https://user-images.githubusercontent.com/30625116/80993710-cad97c00-8e50-11ea-9340-c044d89bf8bf.png)

It will reduce the clutter of the project.

**Describe the solution you'd like**
There is already a way to create an event scriptable object. `Create` button in the variable should create event nested in that variable.

mnarimani:
I created a quick implementation and I'm already loving it. I noticed that `Create` button is drawn with property drawer. So I added additional buttons. But I think the buttons need better placement. Any ideas?
![image](https://user-images.githubusercontent.com/30625116/80997591-c6b05d00-8e56-11ea-9fd0-8e4239714cb6.png)

soraphis:
I thought the same thing, too.
It also solves the problem that created Events are placed in the Assets/ folder which can get kinda annoying

AdamRamberg:
Thanks for this issue @mnarimani! I think this is a great idea. 

We need to make sure that it is possible to both nest and un-nest Events to Variables. My thoughts on this: 
- Just default to always nest Events created via the Inspector.
- If an Event is nested, add a button to un-nest the Event from the Variable. 
- If an Event is not nested to a Variable, then in the Event's editor, add an ObjectField that takes a Variable + add a button to nest the Variable to the Event. 
- Name this feature either: "Create Molecule" or "Fuse / Diffuse". 

@mnarimani @soraphis What are your thoughts on my suggestions? 

---

Furthermore, tested this myself using the following code snippet in `AtomDrawer.cs`:

```
try
{
    // Create asset
    T so = ScriptableObject.CreateInstance<T>();

    // Assign the newly created SO
    property.objectReferenceValue = so;

    if (property.GetParent() is AtomBaseVariable ab)
    {
        so.name = drawerData.NameOfNewAtom;
        AssetDatabase.AddObjectToAsset(so, ab);
        AssetDatabase.SaveAssets();
        AssetDatabase.ImportAsset(AssetDatabase.GetAssetPath(ab));
[...]

PR #316, opened 2021-11-29
#149 [FEATURE] Create button in variables create events as nested scriptable objects (initial implementation).

ThimoDEV:
This PR adds the option to Fuse and Diffuse Atom Events to Atom variables, valueLists, instancers etc. The current implementation is shown in the video's below. NOTE: In the current design its up to the user to configure their project files the way they want. This can lead to possibly unwanted or weird situations, but thats up to the user right now.

Features:

- Fuse/Diffuse referenced events in the Drawer
- Create new standard nested Events
-  Nest ScriptableObjects (and other type of project assets) manually by pressing ALT and dragging the asset onto another asset 
    in the project view.
 
Video 1: https://www.youtube.com/watch?v=6KaW-NeFDTg 

Video 2: https://user-images.githubusercontent.com/68739517/143941070-8175ed01-6bb7-48f7-a66d-3b5faf18fc7c.mp4

### DIFF
--- a/.vsconfig
+++ b/.vsconfig
@@ -1,0 +1,6 @@
+﻿{
+  "version": "1.0",
+  "components": [
+    "Microsoft.VisualStudio.Workload.ManagedGame"
+  ]
+}
--- a/Packages/BaseAtoms/Editor/SubAssetDragAndDrop.cs
+++ b/Packages/BaseAtoms/Editor/SubAssetDragAndDrop.cs
@@ -6,7 +6,7 @@ using UnityEngine;
 
 namespace UnityAtoms.Editor
 {
-    //code found at: https://github.com/Maligan/unity-subassets-drag-and-drop/blob/master/SubAssetDragAndDrop.cs
+    //Code found at: https://github.com/Maligan/unity-subassets-drag-and-drop/blob/master/SubAssetDragAndDrop.cs and modified.
     [InitializeOnLoad]
     public static class SubAssetDragAndDrop
     {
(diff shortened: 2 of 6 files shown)


soraphis:
I think if a reference is removed ... the old value should be checked, if it is fused and automatically defuse.

ThimoDEV:
> I think if a reference is removed ... the old value should be checked, if it is fused and automatically defuse.

Is what you mean the feature that is shown in the first video around 00:35  seconds?

soraphis:
> Is what you mean the feature that is shown in the first video around 00:35 seconds?

ahhh. yes indeed. i had not payed enough attention to it the first time and thought you had clicked defuse.
then i saw in the second video the reference staying in there. at about 00:12

ThimoDEV:
> > Is what you mean the feature that is shown in the first video around 00:35 seconds?
> 
> ahhh. yes indeed (does it also work when overriden?). i had not payed enough attention to it the first time and thought you had clicked defuse. then i saw in the second video the reference staying in there. at about 00:12 when overridden

About your overriden question, I added an option when the user has an Atom fused and drags a new event into the field. The old fused Atom will get diffused and the new reference gets immediately fused. Is this what you mean? An example is shown here: https://youtu.be/wlq2dr1-na0

ThimoDEV:
I would like to suggest to target this PR to the V4.4.x release as I've just sent a PR for this feature targeted at V5. V5 required some additional changes to the AtomDrawer file overall.

soraphis:
> I would like to suggest to target this PR to the V4.4.x release ....

Some time ago it was decided that 4.x won't receive new features but only bug fixes, hence those features target V5

Issue #67, opened 2019-10-20
[FEATURE] Set event name suggestion on Variable Changed event creation


mutmedia:
Is would be really nice to have an automatically set up name for the event when creating a new event for a variable (could be fancy and try to understand what naming convention is being used or just force some naming convention such as appending " changed" or "_changed", or some middle ground such as a setting regex for it?)

This makes the process of creating new events for variables less repetitive (especially in the case mentioned in #66 )

AdamRamberg:
Great idea! I think that just adding the postfix "Changed" to the Event as the default name will suffice for now or what do you think @mutmedia? The tricky part implementing this would be to get the parent of of the `SerializedProperty` in the `AtomDrawer`. You would then use the name of the parent Atom (if there is one) + the name of the property of the Atom you are creating. Eg: if you have a Variable called `MyVariable` and you click Create for the `Changed` event of that Variable the default name would be `MyVariableChanged`.

soraphis:
(untested):
in the event drawer, you can get property.serializedObject.Name which should yield the Scripts's name or the scriptableObjects's asset name.

and of course you could get the property.Name to get the variable name itself.

_________________________________

**scenario 1**: you're in an `IntVariable` AtomVariable named "fooBar" you click "create" for the "changed" event of type IntEvent.

** scenario 2*: you're on an GameObject Player in a Script named Health you click "create" for the "reachedZero" event of type BoolEvent.

| pattern | scenario 1 | scenario 2 |
| <parent>On<Event> | IntVariableOnChanged | HealthOnReachedZero |
| <parent>Changed  | IntVariableChanged | HealthChanged |
| <object name>Changed  | fooBarChanged | PlayerChanged |
| <object name><parent>Changed  | IntVariablefooBarChanged | PlayerHealthChanged |
| <object name><parent>On<Event>  | IntVariablefooBarOnChanged | PlayerHealthOnReachedZero |

AdamRamberg:
I think pattern 3 (row 3) works best for **scenario 1**, however it isn't great for scenario 2. Maybe we could via reflection see if the parent is an Atom Variable or not? If it is we use the `<object name>Changed` pattern, otherwise we just default to `AtomEvent` or maybe `<property name>`?

AdamRamberg:
Fixed in #317

PR #317, opened 2021-12-02
#67 [FEATURE] Set event name suggestion on Variable Changed event creation 

ThimoDEV:
https://youtu.be/_seSt3tMZdU 

Currently the implementation suggests a name for a new Event or variable when created with the 'Create' button. If the parent of the property is a BaseAtom (Variable, valuelist, FSM etc) the naming structure is: "name of variable/valuelist/FSM" + name of property ("Changed" when you add a variable changed event) + type of Event/variable ("BoolEvent" for example). When you try to create a second default named event the system will add a number at the end just like it is with duplicating GameObjects. Everything can be seen in the video above. 

There are some occasions where this default naming scheme can be a bit weird, as you can see with SetIntVariableValue as an example. However if you create custom names for a variable things get quite clear. If there are suggestions for a better naming scheme I will be happy too change things where needed.

### DIFF
--- a/Packages/Core/Editor/Drawers/AtomDrawer.cs
+++ b/Packages/Core/Editor/Drawers/AtomDrawer.cs
@@ -154,18 +154,9 @@ namespace UnityAtoms.Editor
                         drawerData.UserClickedToCreateAtom = true;
                         EditorGUI.FocusTextInControl(NAMING_FIELD_CONTROL_NAME);
 
-                        if (property.GetParent() is BaseAtom abc)
-                        {
-                            var atomName = AtomNameUtils.CheckForDuplicateAtom(abc.name + AtomNameUtils.CleanPropertyName(property.name)
-                                + AtomNameUtils.FilterLastIndexOf(typeof(T).ToString(), "."));
-                            drawerData.NameOfNewAtom = atomName;
-                        }
-                        else
-                        {
-                            var atomName = AtomNameUtils.CheckForDuplicateAtom(AtomNameUtils.CleanPropertyName(property.name)
-                                + AtomNameUtils.FilterLastIndexOf(typeof(T).ToString(), "."));
-                            drawerData.NameOfNewAtom = atomName;
-                        }
+                        var baseAtomName = AtomNameUtils.CleanPropertyName(property.name) + typeof(T).Name;
+                        var atomName = property.GetParent() is BaseAtom parentAtom ? parentAtom.name + baseAtomName : baseAtomName;
+                        drawerData.NameOfNewAtom = AtomNameUtils.CreateUniqueName(atomName);
                     }
                 }
             }
--- a/Packages/Core/Editor/Drawers/AtomDrawer.cs
+++ b/Packages/Core/Editor/Drawers/AtomDrawer.cs
@@ -1,6 +1,7 @@
 #if UNITY_2018_3_OR_NEWER
 using System.Collections.Generic;
 using System.IO;
+using System.Text.RegularExpressions;
 using UnityEditor;
 using UnityEngine;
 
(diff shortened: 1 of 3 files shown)


miikalo:
Regarding 53686d8, you don't need to worry about supporting Unity 2019, feel free to use the 2020+ features.

ThimoDEV:
I would like to suggest to target this PR to the V4.4.x release as I've just sent a PR for this feature targeted at V5. V5 required some additional changes to the AtomDrawer file overall.

Packages/Core/Editor/Drawers/AtomDrawer.cs
```diff
@@ -151,10 +151,21 @@ public override void OnGUI(Rect position, SerializedProperty property, GUIConten
...
+                            var atomName = AtomNameUtils.CheckForDuplicateAtom(abc.name + AtomNameUtils.CleanPropertyName(property.name)
+                                + AtomNameUtils.FilterLastIndexOf(typeof(T).ToString(), "."));
+                            drawerData.NameOfNewAtom = atomName;
+                        }
+                        else
+                        {
+                            var atomName = AtomNameUtils.CheckForDuplicateAtom(AtomNameUtils.CleanPropertyName(property.name)
+                                + AtomNameUtils.FilterLastIndexOf(typeof(T).ToString(), "."));
+                            drawerData.NameOfNewAtom = atomName;
+                        }
```
AdamRamberg, line 168: This could be rewritten with less repetion like this: 
```
                        var baseAtomName = AtomNameUtils.CleanPropertyName(property.name) + AtomNameUtils.FilterLastIndexOf(typeof(T).ToString(), ".");
                        var atomName = property.GetParent() is BaseAtom parentAtom ? parentAtom.name + baseAtomName : baseAtomName;
                        drawerData.NameOfNewAtom = AtomNameUtils.CheckForDuplicateAtom(atomName);
```

Packages/Core/Editor/Utils/AtomNameUtils.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+            if (Regex.Match(cleanedProperty, @"[a-zA-Z]").Success)
+            {
+                var index = Regex.Match(cleanedProperty, @"[a-zA-Z]").Index;
+                cleanedProperty = cleanedProperty[index].ToString().ToUpper() + cleanedProperty.Substring(index + 1);
+            }
+            return cleanedProperty;
+        }
+
+        public static string CheckForDuplicateAtom(string atomName)
```
AdamRamberg, line 28: It is not really checking for duplicates, but rather giving the atom an unique name based on if there are other assets with the same name. Could we call it `CreateUniqueName` instead?

AdamRamberg (changes requested):
Very nice PR! Got a couple of minor things to look at. If you do the changes I will then approve the PR and merge to `canary`.

Packages/Core/Editor/Utils/AtomNameUtils.cs
fakegood: there's utility functions from Unity, [GenerateUniqueAssetPath](https://docs.unity3d.com/ScriptReference/AssetDatabase.GenerateUniqueAssetPath.html), and [GetUniqueName](https://docs.unity3d.com/ScriptReference/ObjectNames.GetUniqueName.html) can be considered for this use case.

Packages/Core/Editor/Drawers/AtomDrawer.cs
AdamRamberg: Fixed in latest commit

Packages/Core/Editor/Utils/AtomNameUtils.cs
AdamRamberg: @fakegood Didn't know about those 2, great tip! However, I believe that in this case the utils that Thimo has made are actually easier to use / maintain.

PR #318, opened 2021-12-05
#194 [FEATURE] Replay Buffer Settings 

ThimoDEV:
Added a few things regarding the settings for replay buffers.

- Global Replay buffer default setting (Edit => preferences => Unity Atoms => Default replay buffer)
- Option to use a local replay buffer setting for an event.
- Added ReplayEventBufferOnListener property to change replay listening through code.
- Added replayEventsBuffer as parameter in Register method of Events (default is true). This way when registering through code tthe replay buffer can be ignored.
 
 Note: When testing with the infintewaves example, make sure you set _useLocalReplayBuffer to true to make sure the example scene and scriptableobjects do not get reset to the global default value.
 
 Found in class Packages/Core/Runtime/Events/AtomEvent.cs line 29.

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -25,6 +25,9 @@ namespace UnityAtoms
         [SerializeField]
         protected event Action<T> _onEvent;
 
+        [SerializeField]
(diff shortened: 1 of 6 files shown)


soraphis:

PR #340, opened 2022-03-13
Fix #321 NRE when Event Reference Listeners event reference is not set

### LINKED ISSUE #321: [BUG] NullReferenceException when Event Reference Listeners event reference is not set
Event Reference Listener throws a null reference exception when not using "use event".

the getter for Event in AtomEventReference.cs should be changed, to include null checks:

```
        public override E Event
        {
            get
            {
                switch (_usage)
                {
                    case (AtomEventReferenceUsage.VARIABLE): return _variable.GetOrCreateEvent<E>();
                    case (AtomEventReferenceUsage.VARIABLE_INSTANCER): return _variableInstancer.GetOrCreateEvent<E>();
                    case (AtomEventReferenceUsage.EVENT_INSTANCER): return _eventInstancer.Event;
                    case (AtomEventReferenceUsage.EVENT):
                    default:
                        return _event;
                }
            }
```

like this:

```
                    case (AtomEventReferenceUsage.VARIABLE): return _variable == null ? null : _variable.GetOrCreateEvent<E>();
                    case (AtomEventReferenceUsage.VARIABLE_INSTANCER): return _variableInstancer == null ? null : _variableInstancer.GetOrCreateEvent<E>();
                    case (AtomEventReferenceUsage.EVENT_INSTANCER): return _eventInstancer == null ? null : _eventInstancer.Event;
```

using null could be an option for defining the value in runtime, it should not throw a NRE

ThimoDEV:
As suggested by @soraphis changed that the Event reference listener doesn't just throw a null reference exception everywhere. I added a check for Atoms debug mode. In debug mode the user gets notified of the null reference which he has. But in runtime/production this exception doesn't get thrown.

I chose for a NRE because on clicking this error the connected gameObject gets pinged. This doesn't happen when using the Unity Debug class.

### DIFF
--- a/Packages/Core/Runtime/EventReferences/AtomEventReference.cs
+++ b/Packages/Core/Runtime/EventReferences/AtomEventReference.cs
@@ -27,12 +27,9 @@ namespace UnityAtoms
             {
                 switch (_usage)
                 {
-                    case (AtomEventReferenceUsage.VARIABLE):
-                        return _variable == null ? LogNullReference(nameof(_variable)) : _variable.GetOrCreateEvent<E>();
-                    case (AtomEventReferenceUsage.VARIABLE_INSTANCER):
-                        return _variableInstancer == null ? LogNullReference(nameof(_variableInstancer)) : _variableInstancer.GetOrCreateEvent<E>();
-                    case (AtomEventReferenceUsage.EVENT_INSTANCER):
-                        return _eventInstancer == null ? LogNullReference(nameof(_eventInstancer)) : _eventInstancer.Event;
+                    case (AtomEventReferenceUsage.VARIABLE): return _variable == null ? null : _variable.GetOrCreateEvent<E>();
+                    case (AtomEventReferenceUsage.VARIABLE_INSTANCER): return _variableInstancer == null ? null : _variableInstancer.GetOrCreateEvent<E>();
+                    case (AtomEventReferenceUsage.EVENT_INSTANCER): return _eventInstancer == null ? null : _eventInstancer.Event;
                     case (AtomEventReferenceUsage.EVENT):
                     default:
                         return _event;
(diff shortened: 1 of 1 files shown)


soraphis:
>  This doesn't happen when using the Unity Debug class.

Debug.Log can take a second arguments. you can pass the object to highlight as second parameter.


I don't think it should throw an exception - even in debug mode - because it can be null _ON PURPOSE_ which would mean that your script only works if not in debug mode.

with throwing the NRE in our scripts, we take away the oportunity to null check for users of the library. when returning null it might throw a NRE in the users code - but this isn't much of a problem for the user to fix, he can just check if this getter has returned null and act accordingly.

when throwing a NRE he cannot simply check, but instead is forced to use a try-catch block

ThimoDEV:
_with throwing the NRE in our scripts, we take away the oportunity to null check for users of the library. when returning null it might throw a NRE in the users code - but this isn't much of a problem for the user to fix, he can just check if this getter has returned null and act accordingly._

Ah yes this makes a lot of sense to not choose for throwing an exception. 

The only problem I have right now, I can't find the gameObject that throws the error/log/warning from within AtomEventReference class. Do you have a suggestion where or how I can find this?

soraphis:
> The only problem I have right now, I can't find the gameObject that throws the error/log/warning from within AtomEventReference class. Do you have a suggestion where or how I can find this?

why even log something? ¯\_(ツ)_/¯
the same applies: if the user code want's to check if it is `null` you're now spamming him with this log message in debug mode.
if you remove the logging altogether, a NRE will be thrown in his own code, providing him all information needed.

if you want to supply the developer with additional information about unassigned references it could be more helpful to create an editor extension, that shows either a) in the inspector that a field must be designed to work properly (e.g. 'required' in https://github.com/dbrizov/NaughtyAttributes#required) or b) show somewhere that (and how many) of those game objects with missing required assignments exist in the scene. 🤔

ThimoDEV:
The unassigned refererences part seems quite a useful thing to implement for Atoms in my opinion. Or else users might be searching which of their objects has a missing reference. I think its something for V5 as a nice addition. For now I removed the logging.

soraphis (approved):

miikalo (approved):

PR #362, opened 2022-07-05
fix: corrected IEquatable implementation check

### LINKED ISSUE #361: [BUG] Asset based Generator does ignore IEquatable implementation
**Describe the bug**
When using the asset based generator, the generated files always assume that the type is not an IEquatable

**To Reproduce**
Generate a Int32 AtomVariable, it has a NotImplementedException ValueEquals method.

**Expected behavior**
Since Int32 implements IEquatable<Int32>, it should inherit from EquatableAtomVariable

**Environment**
 - Version of Unity Atoms [4.4.5]

**Additional context**
the problem is this check here: https://github.com/unity-atoms/unity-atoms/blob/5df498b41ee77de0c5faa73b76f7fd0ed890216a/Packages/Core/Editor/Generator/AtomGenerator.cs#L34=

this is never the case

soraphis:
fixes #361


AdamRamberg (approved):
Super nice! <3

Issue #303, opened 2021-09-30
[BUG] Errors when viewing atoms generated for non serializable types


IceTrooper:
**Describe the bug**
Unity Atoms custom editors throw a NullReferenceException when custom generated type is not serializable. It's only the custom Editor windows/Drawers issue.

**To Reproduce**
For example:
1. Generate custom Atom for TimeSpan
2. Create TimeSpanVariable
3. Focus TimeSpanVariable
4. Many NullReferenceException errors appear in a console, because of the custom Inspector.
5. Similar errors are thrown when creating AtomReference or AtomVariable fields in custom scripts.

**Expected behavior**
Non-serializable types should work with UnityAtoms. Of course, we can't display values in the Editor because of the Unity serialization system, but... we should still be able to create and use those types of Atoms via code without any restrictions or making wrappers.

**Additional context**
https://discord.com/channels/640589221136171020/641007735525081116/893103547271487529
Copy-pasted from discord:
> You're right that I won't see any non-serializable type... and it's the desired effect. That's the case. I want to use AtomVariable/Events/etc. based on non-serialized type and I should not care about viewing those values in the Editor. I want to use it only for a scripting communication process, not for a designer. 
The problem is that UnityAtoms throws errors for non-serializable types when you focus on them. However, I think this should not be the case. Instead, it should just show things it can display like Developer Description, Id, Changed, etc. and not throwing errors.
In my case, I don't want to see the TimeSpan value in the Editor. I only want not to have errors in my console when I click on that Variable. The wrapper for TimeSpan is not the real solution in my opinion, because devs shouldn't be forced to make a wrapper if they don't care about viewing values in the editor in the first place. 
I can offer a pull request as I correct these errors. I want to have your opinion about it.

AdamRamberg:
Fixed in #371

PR #369, opened 2022-10-18
removed ondisable method, to fix #349

### LINKED ISSUE #349: [BUG] Default changed and changedWithHistory events override custom events
**Describe the bug**
The fix for the problem when default changed and changedWithHistory events were not created introduced  the bug which overrides the current change event when a change event is added to a variable after running with empty change events.

**To Reproduce**

1.  Leave variable changed event blank and in runtime change variable value to something. Stop runtime.
2.  Add your change event to the variable, and go into runtime again and change variable value. Check variable event, it changes to default.

**Environment**
 - OS: [Windows 10]
 - Unity Version [2020.3.25f1]
 - Version of Unity Atoms [v4.4.4]

soraphis:
this addresses #349 and fixes an issue introduced with #303 

when starting the runtime without changed event, it would set the "_changedInstantiatedAtRuntime" to true.

OnDisable is only called if you hit play again and are about to enter the playmode again, which then RESETS all your changes (e.g. assigning an event in edit time) to null. 

the alternative would be to address this in the custom editor drawer, to set _changedInstantiatedAtRuntime to false again when _changed was set.

but I found that the OnDisable was not needed, as the values where already set back when leaving playmode and in the build (which i have not tested!! it wouldn't matter anyways since OnDisable is only called if the memory is freed altogether)

### DIFF
--- a/Packages/Core/Runtime/Variables/AtomVariable.cs
+++ b/Packages/Core/Runtime/Variables/AtomVariable.cs
@@ -47,7 +47,6 @@ namespace UnityAtoms
         [SerializeField]
         [FormerlySerializedAs("Changed")]
         private E1 _changed;
-        private bool _changedInstantiatedAtRuntime;
         public E1 Changed
         {
             get
@@ -67,7 +66,6 @@ namespace UnityAtoms
         [SerializeField]
         [FormerlySerializedAs("ChangedWithHistory")]
         private E2 _changedWithHistory;
-        private bool _changedWithHistoryInstantiatedAtRuntime;
         public E2 ChangedWithHistory
         {
             get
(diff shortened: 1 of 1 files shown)


AdamRamberg:
Looks good to me ✅ Pushed a small commit removing `_changed` and `_changedInstantiatedAtRuntime` since they are not used anymore after removing `OnDisable`.

AdamRamberg (approved):

soraphis:
@AdamRamberg @soraphis 
not sure anymore if this will work.

Should do a check with disables domain reload and stuff.

AdamRamberg:
@soraphis Not sure I follow. What exactly do you think will break in this change? Do you think we should revert this before doing the next release?

soraphis:
When i find time again, i'll need to recheck again if everything is correctly resetted.

I think if domain reload is off, neither onEnable nor onDisable are called.

So this should not break anything, but i wanna double check state is correctly cleared between playmode switches

AdamRamberg:
Cool! Lets wait with the release until that then

soraphis:
ok, as far as i can tell it still works correctly even if domain+scene reload is disabled in the editor settings.
so release is good to go imho. 👍

PR #304, opened 2021-09-30
#303 fixed custom editor errors for non-serializable types

IceTrooper:
PR for https://github.com/unity-atoms/unity-atoms/issues/303

PR #371, opened 2022-10-23
prevent null reference exceptions in editor when using not-serializab…

soraphis:
this is taking only the relevant parts of PR #304. Special thanks to @IceTrooper.

I created a new PR to keep all the small refactoring changes out, which resulted in merge conflicts and much noise.

In AtomBaseReferenceDrawer > OnGUI: the diff looks like much, but I actually just wrapped the previous code in an if-else branch and I'm showing a label if the property is not serializable. the else branch is unchanged.

I suggest in the "files changed"-tab select the settings-icon and "hide whitespaces" so the indentation does not confuse the diff coloring

### DIFF
--- a/Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs
+++ b/Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs
@@ -59,6 +59,7 @@ namespace UnityAtoms.Editor
             // Calculate rect for configuration button
             Rect buttonRect = new Rect(position);
             buttonRect.yMin += _popupStyle.margin.top;
+            buttonRect.yMax = buttonRect.yMin + EditorGUIUtility.singleLineHeight;
             buttonRect.width = _popupStyle.fixedWidth + _popupStyle.margin.right;
             position.xMin = buttonRect.xMax;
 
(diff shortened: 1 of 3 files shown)


Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs
```diff
@@ -56,7 +59,6 @@ public override void OnGUI(Rect position, SerializedProperty property, GUIConten
             // Calculate rect for configuration button
             Rect buttonRect = new Rect(position);
             buttonRect.yMin += _popupStyle.margin.top;
-            buttonRect.yMax = buttonRect.yMin + EditorGUIUtility.singleLineHeight;
```
AdamRamberg, line 59: Removing this line is a mistake right?

AdamRamberg (changes requested):

Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs
AdamRamberg: Reverted in last commit

AdamRamberg (approved):
Super nice! Thanks for submitting this PR <3

PR #386, opened 2023-01-29
FiniteStateMachineMonoHook unload fix

### LINKED ISSUE #385: FiniteStateMachineMonoHook object gets created in scene with built index "-1"
https://github.com/unity-atoms/unity-atoms/blob/5df498b41ee77de0c5faa73b76f7fd0ed890216a/Packages/FSM/Runtime/FiniteStateMachine/FiniteStateMachine.cs#L84

This line is called when the fsm (which is an asset) is loaded, at this point, the first scene might be not loaded yet. (SceneManager.GetActiveScene() returns a unnamed scene with buildIndex: -1)

So when the scene gets loaded, the object is Destroyed. And the FSM is not updated.

soraphis:
prevents FiniteStateMachineMonoHook singleton from being unloaded

fixes #385


AdamRamberg (approved):
LGTM ✅

PR #389, opened 2023-02-05
fixes syncgameobjecttolist adding object multiple times

### LINKED ISSUE #375: [BUG] `SyncGameObjectToList` add the game object multiple times
**Describe the bug**
Component `SyncGameObjectToList` is pretty intuitive in its purpose, but it happens to add the game object multiple times to the referred list, if the object is enabled and disabled at runtime.

**To Reproduce**
1. Add any game object to the scene;
2. Add the component `SyncGameObjectToList`;
3. Fill its `list` field with a valid variable;
4. Play the scene;
5. In inspector, activate and deactivate the GameObject repeteadly (you can do it by code, if you prefer).
6. Observe the game object is added again and again to the list "OnEnable".

**Expected behavior**
The game object instance should be added to the list once, instead of multiple times. I don't know why `OnEnable`/`OnDestroy` pair were used instead of `Awake`/`Destroy`, but it can fit there too with a simple "contains" check. It would be useful as well if it could handle cases where the game object is removed on disable, because *sometimes* (not always) the developer may want to remove it from the list if disabled. Could it be a bool field?

**Environment**
 - OS: Windows
 - Unity Version: 2021.3.3f1
 - Version of Unity Atoms: 4.4.5

**Additional context**

Line causing the issue:

https://github.com/unity-atoms/unity-atoms/blob/5df498b41ee77de0c5faa73b76f7fd0ed890216a/Packages/BaseAtoms/Runtime/MonoBehaviourHelpers/SyncGameObjectToList.cs#L19

### LINKED ISSUE #388: [Bug] SyncGameObjectToList adds object multiple times to list
https://github.com/unity-atoms/unity-atoms/blob/5df498b41ee77de0c5faa73b76f7fd0ed890216a/Packages/BaseAtoms/Runtime/MonoBehaviourHelpers/SyncGameObjectToList.cs#L16-L25

OnEnable should partner with OnDisable.

OnDestroy should partner with Start.

other Sync* classes use Start/OnDestroy pair, so we should do this here as well

soraphis:
fixes #388
fixes #375

### DIFF
--- a/Packages/BaseAtoms/Runtime/MonoBehaviourHelpers/SyncGameObjectToList.cs
+++ b/Packages/BaseAtoms/Runtime/MonoBehaviourHelpers/SyncGameObjectToList.cs
@@ -4,7 +4,7 @@ using UnityEngine.Assertions;
 namespace UnityAtoms.BaseAtoms
 {
     /// <summary>
-    /// Adds a GameObject to a GameObject Value List on OnEnable and removes it on OnDestroy.
+    /// Adds a GameObject to a GameObject Value List on Start and removes it on OnDestroy.
     /// </summary>
     [AddComponentMenu("Unity Atoms/MonoBehaviour Helpers/Sync GameObject To List")]
     [EditorIcon("atom-icon-delicate")]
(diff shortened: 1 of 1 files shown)


AdamRamberg (approved):
Thanks for submitting this PR! 🙏 LGTM ✅

PR #403, opened 2023-05-26
fix: Event Replay Buffer not cleared when domain reload disabled

### LINKED ISSUE #146: [Feature] Making UnityAtoms compatible with Enter The PlayMode feature of Unity
Enter The PlayMode is a feature that makes entering the play mode insanely fast. 
image

But there are side effects:

- Fields marked with [NonSerialized] attribute keep their values - as script serialization step is skipped completely and scripts remain the same, all data is carried to the Play Mode.
- Static variables keep their values - because Domain Reload does not happen types stay initialized.
- Static events keep their subscribers.
- There is no extra OnDisable/OnEnable calls for ExecuteInEditMode scripts.
- ScriptableObject, MonoBehaviour fields which are not serialized into the build ([NonSerialized], private or internal) keep their values - as existing objects are not recreated and constructors are not called.
- Watch out for null private and internal fields of array/List type - they are converted to an empty array/List object during Domain Reload and stay non null for game scripts.
- ExecuteInEditMode/ExecuteAlways scripts are not destroyed/awaken - no OnDestroy/Awake calls for those.
- Watch out for Awake/OnEnable methods which check EditorApplication.isPlaying property - Awake is not called and OnEnable is called only when EditorApplication.isPlaying is already true on Play Mode change.

UnityAtoms gets heavily impacted by these side effects.
**OnEnable and OnDisable calls are not called anymore.** So nothing works as expected.

I exactly don't know how we are supposed to change the code at the moment. So I created this issue for the discussion about it.

There are some useful callbacks that we can use. 

- `EditorApplication.playModeStateChanged` event
- ` [RuntimeInitializeOnLoadMethod]` attribute

What do you think?

soraphis:
this fixes a discrepancy between the AtomVariable and AtomEvent implementation.

While the AtomVariable handled disabled domain reload already (since 4.4.) AtomEvent still had their replay buffer persisting.
This was first mentioned in https://github.com/unity-atoms/unity-atoms/issues/146#issuecomment-813027910

steps to reproduce:
- disable domain reload
- add a variable to a listener and attach an action that logs the value.
- hit play, invoke the event.
- hit stop, and play again

result without this commit: the last value is logged. with this commit, the buffer is cleared correctly and no value is logged.

The implementation follows the one in AtomVariable.

fixes #146

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -1,6 +1,9 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
 using UnityEngine;
 
 namespace UnityAtoms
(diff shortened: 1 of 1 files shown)


soraphis:
Only thing i wonder is, if we should also set '_onEvent' to null ... (effectively deregistering all listeners) in the same method...

AdamRamberg (approved):
Looks good to me! 

Regarding your question, I don't think we should clear `_onEvent`, but haven't tested it and I'm not 100% familiar with all the ins and outs of enter play mode. What do you think could be the issue if we don't clear it?

soraphis:
> What do you think could be the issue if we don't clear it?

I have not tested it, so I can't say if the following is realistically what would happen:

some custom component registers in OnEnable - but does not Unregister.
Now the event keeps the C# part of the component alive, while the native object below is destroyed. Now "raise" is called either from the editor button or in the next playmode session...

AdamRamberg:
Ah, I see. I believe we should clear `_onEvent` then. My suggestion is to use `UnregisterAll()`. Shouldn't we, for consistency sake, also call `UnregisterAll()` in `OnDisable()` instead of this: 
```csharp
            // Clear all delegates when exiting play mode
            if (_onEvent != null)
            {
                var invocationList = _onEvent.GetInvocationList();
                foreach (var d in invocationList)
                {
                    _onEvent -= (Action<T>)d;
                }
            }
```

AdamRamberg (changes requested):
See my last comment

soraphis:
> If you implement the above and it seems to work fine, then I think we are good to merge.
well, the OnDisable code is quite old :D https://github.com/unity-atoms/unity-atoms/pull/114

I agree clearing _onEvent in OnDisable / OnEnable by calling UnregisterAll should be the way to go.

AdamRamberg (approved):
Looks great ✅

PR #407, opened 2023-07-07
fix: IEquatable actually checks for the generic interface

RedGlow:
This pull request fixes the check for the type to implement the `System.IEquatable<>` interface in the (asset based?) generator.

Currently the check is wrong, and always results in the type not being equatable, because `IEquatable` is a generic type, and thus requires a slightly different approach (see also [this StackOverflow answer](https://stackoverflow.com/questions/4963160/how-to-determine-if-a-type-implements-an-interface-with-c-sharp-reflection/4963190#4963190)).


soraphis:
duplicate of https://github.com/unity-atoms/unity-atoms/pull/362 ?

RedGlow:
Yes, I think this fixes the same bug. I don't know honestly if there could be a difference between your fix and mine. Maybe IsAssignableFrom is more generic (one could derive from an interface that derives on its own from IEquatable)? In this case the currently existing one is better.

PR #409, opened 2023-07-12
FIX: Using TextField in AssetGenerator does not work as expected

### LINKED ISSUE #408: [BUG] Generator asset fails to generate some `UnityEngine` types.
**Describe the bug**
Trying to generate types from the `UnityEngine` namespace doesn't work, or provides weird results. For example, in the case of `UnityEngine.Camera` some variable scripts are generated, but they are missing the _Camera_ prefix. See screenshot below. In the case of `UnityEngine.Grid` it generates the subclass, `GridLayout`. 

**To Reproduce**
1. Create an `Atom Generator` asset
2. Paste this line into the `Full Qualified Name` field
_UnityEngine.Camera, UnityEngine, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null_
or
_UnityEngine.Grid, UnityEngine, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null_

**Expected behavior**
Unity types are able to be generated via the `Atom Generator` asset

**Screenshots**
Results
image

**Environment**
 - OS: Windows 11
 - Unity Version: 2021.3.24f1
 - Version of Unity Atoms: `canary` branch

soraphis:
fix: pasting FullQualifiedName in the generators text field, not picking up the actual type

closes #408

### DIFF
--- a/Packages/Core/Editor/Editors/AtomGeneratorEditor.cs
+++ b/Packages/Core/Editor/Editors/AtomGeneratorEditor.cs
@@ -245,7 +245,11 @@ namespace UnityAtoms.Editor
                         parent.AddChild(dropdownItem);
 
                         // Use Hash instead of id! If 2 AdvancedDropdownItems have the same name, they will generate the same id (stupid, I know). To ensure searching for a unique identifier, we use the hash instead.
-                        if (!idTypePairs.TryAdd(dropdownItem.GetHashCode(), type) && type.FullName != idTypePairs[dropdownItem.GetHashCode()].FullName)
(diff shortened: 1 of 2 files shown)


AdamRamberg (changes requested):
Fix looks good! However, `TryAdd` doesn't exist in older versions of .NET like .NET 4 that we use in `unity-atoms-examples`. See error: 
Skärmavbild 2023-07-18 kl  23 01 03

Fix this by creating an extension or do it inline with `ContainsKey` + `Add`. When fixed this one is ok to merge.

AdamRamberg (approved):
Looks great 🚢

PR #413, opened 2023-07-19
Adding Documentation

### LINKED ISSUE #410: [Documentation] Add section about Pre Change Transformers
The current documentation does not mention or explain the usage of "Pre Change Transformers"

soraphis:
fixes #410 
I've added an own "philosophy" page and made the "overview" way slimmer - and hopefully less intimidating. It's now just mentioning all (or at least most of) the features and relying more on them having their own tutorial page.

also added a page for the Tags package.

Note, that I made heavy use of ChatGPT 🙈 - but I've added a disclaimer on all relevant pages, the disclaimer might not be styled correctly though. I've used the [Admonitions](https://docusaurus.io/docs/markdown-features/admonitions) styling but it is docusaurus version 2.0 and were still using v 1.x i think


AdamRamberg:
Prefering this PR over #416. I also made the following changes:

* Removed everything that talks about implementations in the Philopsophy section.
* Added notes from Ryan Hipple's talk to the philosophy section
* Reintroduced the old overview page with changes from https://github.com/unity-atoms/unity-atoms/pull/416, since I believe it is
* Removed all Admonitions, since they are not working in docusaurus v1
* Proof read and made corrections to variables transformers and tags pages.
* Added changes in "creating atoms" from https://github.com/unity-atoms/unity-atoms/pull/416

Issue #412, opened 2023-07-19
[FEATURE] Tag Package, Events


soraphis:
The Tags subpackage should come with events that are triggered when tags are added or removed:

- `OnTagAdded`
- `OnTagRemoved`

ThaungThanHan:
I would like to work on this feature. Could you please assign it to me?

PR #415, opened 2023-07-20
feature: Added OnTagAdded and OnTagRemoved event handlers

ThaungThanHan:
In this PR, I have added OnTagAdded and OnTagRemoved event handlers. Those two functions will fire whenever a tag is added or removed. In this procedure, I created a tagEventSubscriber script to connect the event listener to the AtomTags script in Tags package. Please review the code and feel free to pinpoint the area I am missing out on.
Fixes: #412


soraphis:
please make yourself familiar with the https://github.com/unity-atoms/unity-atoms/blob/master/CONTRIBUTING.md 

pull requests should target the canary branch, not the master.

Packages/Tags/Runtime/TagsEventsSub.cs
```diff
@@ -0,0 +1,36 @@
+using UnityEngine;
+using UnityAtoms.Tags;
+
+public class TagEventSub : MonoBehaviour
```
soraphis, line 4: This should not be in the repository, this is just for testing purposes, either have it in an example project or create an actual test case in in Packages/Tags/Test not Runtime...

Packages/Tags/Runtime/AtomTags.cs
```diff
@@ -16,6 +16,9 @@ namespace UnityAtoms.Tags
     [AddComponentMenu("Unity Atoms/Tags")]
     public sealed class AtomTags : MonoBehaviour, ISerializationCallbackReceiver
     {
+        // declaring events
+        public event Action<string> TagAdded;
+        public event Action<string> TagRemoved;
```
soraphis, line 21: I think It'd be more ideomatic to use AtomEvents here...

soraphis (changes requested):

Packages/Tags/Runtime/TagsEventsSub.cs
ThaungThanHan: I have moved it to Packages/Tags/Test.

Packages/Tags/Runtime/AtomTags.cs
ThaungThanHan: I have replaced with using AtomEvents there.

Packages/Tags/Runtime/TagsEventsSub.cs
soraphis: Thats not what I meant. In the Test assembly, there should be tests, that are actually executable by the unity testrunner.

Your code is a MonoBehaviour, you could write a testcase using that behaviour, or create a PR on the https://github.com/unity-atoms/unity-atoms-examples repository, where you're showing of the usage. 

Or write a page of documentation for the feature, where you show of the usage.

But on it's own, this class should not be in the repository, as it does not provide anything to it.

So, to summarize:
- either remove it from the PR (you can keep it locally, though)
- or create unit tests that make use of the class.

dkasfn
soraphis, line 1: what is this file doing here?

ThaungThanHan: Sorry, I have deleted it. I think it occurred when I rebased.

Packages/Tags/Runtime/TagsEventsSub.cs
ThaungThanHan: No worries. Now that I understood what you meant, I have removed it.

soraphis (changes requested):
Could you try to get rid of the changes in docs/assets/mono-hooks/listener.png and docs/tutorials/mono-hooks.md ?
I think they come from the fact, that you developed on master, instead of canary.

you can probably resolve that via

`git checkout origin/canary -- docs/assets/mono-hooks/listener.png`
`git checkout origin/canary -- docs/tutorials/mono-hooks.md`

this will revert the file to the state it has on the origin/canary.

Packages/Tags/Runtime/AtomTags.cs
```diff
@@ -16,6 +16,9 @@ namespace UnityAtoms.Tags
     [AddComponentMenu("Unity Atoms/Tags")]
     public sealed class AtomTags : MonoBehaviour, ISerializationCallbackReceiver
     {
+        // declaring events
+        public AtomEventString OnTagAdded;
+        public AtomEventString OnTagRemoved;
```
soraphis, line 21: Where does the `AtomEventString` type come from? I can't find it in the project.
It should be `StringEvent`, shouldn't it?

https://github.com/unity-atoms/unity-atoms/blob/canary/Packages/BaseAtoms/Runtime/Events/StringEvent.cs

Packages/Tags/Runtime/AtomTags.cs
```diff
@@ -172,6 +175,9 @@ public void AddTag(StringConstant tag)
             // Update static accessors:
             if (!TaggedGameObjects.ContainsKey(tag.Value)) TaggedGameObjects.Add(tag.Value, new List<GameObject>());
             TaggedGameObjects[tag.Value].Add(this.gameObject);
+
+            // Start TagAdded event
+            TagAdded?.Invoke(tag.Value);
```
soraphis, line 180: When using StringEvent, this will be `Raise` instead of `Invoke`.
Also StringEvent is a UnityEngine.Object type, so it must not be compared with `?.` but instead using `if(... == null)` because the UnityEngine.Object `==` operator is overloaden and this is not taken into account when using `?.` or `is null`

soraphis (changes requested):

PR #416, opened 2023-07-20
Docs improvemens: Overivew renaming and pre change transformers

AdamRamberg:
This PR addresses #410. 

Changes: 
- Rename the section "Overview and philosopy" to just "Overview"
- Add docs section on pre change transformer in the variables tutorial
- Minor nit fixes


soraphis:
:D I also tried to address #410 
see: https://github.com/unity-atoms/unity-atoms/pull/413

I first thought of taking the best of both worlds, I like that you added pictures (as I did not), but my changes are a good amount larger, so it's not super easy to take both changesets.

I reviewed your changes and have nothing against it, but I will wait with merging until you took a look onto my changes, maybe you find a thing or two you like.

e.g. I created a whole page for pre-change-transformers, and added a code example to how to create one on your own (sure, there is already the ClampInt provided by the library and people could look into that, but i feel people would more likely look into the documentation than into the source code)

also, I added a philosophy page and a page explaining tags.



I feel we could probably take both PRs. E.g. starting with yours and then taking mine where applicable (or vice versa)

AdamRamberg:
Prefering #413

Issue #419, opened 2023-07-23
[BUG] AtomValueList not compatible with disabled domain reload


soraphis:
relating to #146

currently Variables (https://github.com/unity-atoms/unity-atoms/blob/canary/Packages/Core/Runtime/Variables/AtomVariable.cs#L193-L209) and Events (https://github.com/unity-atoms/unity-atoms/blob/canary/Packages/Core/Runtime/Events/AtomEvent.cs#L63-L81) are handling the disabled domain reload fine, but AtomValueLists don't

e.g. it's not cleared on enter play mode even if "startCleared" is checked (see also #418 )

Issue #418, opened 2023-07-23
[BUG] AtomValueList keeps runtime modifications


soraphis:
**Describe the bug**
Since AtomValueLists don't have an "initial" field (like variables have), they will keep runtime modifications after exiting playmode in the editor. 

(obviously "start cleared" does not suffer this issue (with enabled domain reload, for disabled see #419))

**Environment**
 - Unity Version [e.g. 2019.3]
 - Version of Unity Atoms [4.4.6]

AdamRamberg:
Fixed in #420

PR #420, opened 2023-07-24
AtomValueList disabled domain reload compatibility

soraphis:
Fixes: #419 Fixes: #418 
Minor fix, for an inaccuracy of #403

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -62,7 +62,7 @@ namespace UnityAtoms
 #if UNITY_EDITOR
         private static void HandlePlayModeStateChange(PlayModeStateChange state)
         {
-            if (state == PlayModeStateChange.ExitingEditMode)
+            if (state == PlayModeStateChange.ExitingEditMode) // BEFORE any GO is initialized:
             {
                 foreach (var instance in _instances)
                 {
@@ -70,7 +70,7 @@ namespace UnityAtoms
                     instance.UnregisterAll();
                 }
             }
-            else if (state == PlayModeStateChange.EnteredPlayMode)
+            else if (state == PlayModeStateChange.EnteredEditMode) // AFTER Playmode stopped
             {
                 foreach (var instance in _instances)
                 {
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -62,15 +62,8 @@ namespace UnityAtoms
 #if UNITY_EDITOR
         private static void HandlePlayModeStateChange(PlayModeStateChange state)
         {
-            if (state == PlayModeStateChange.ExitingEditMode) // BEFORE any GO is initialized:
-            {
-                foreach (var instance in _instances)
-                {
-                    instance._replayBuffer.Clear();
-                    instance.UnregisterAll();
-                }
-            }
-            else if (state == PlayModeStateChange.EnteredEditMode) // AFTER Playmode stopped
+            if (state == PlayModeStateChange.ExitingEditMode // BEFORE any GO is initialized:
+                || state == PlayModeStateChange.EnteredEditMode) // AFTER Playmode stopped
             {
                 foreach (var instance in _instances)
                 {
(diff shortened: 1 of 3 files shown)


Packages/Core/Runtime/ValueLists/AtomValueList.cs
```diff
@@ -42,6 +44,59 @@ public abstract class AtomValueList<T, E> : BaseAtomValueList, IList<T>
...
+        {
+            if (state == PlayModeStateChange.ExitingEditMode) // BEFORE any GO is initialized:
+            {
+                foreach (var instance in _instances)
+                {
+                    if (instance == null)
+                    {
+                        _instances.Remove(instance);
+                        continue;
+                    }
```
AdamRamberg, line 81: This could cause an `InvalidOperationException` when removing an item in the set while iterating over it. I get the need though to remove an instance that is no longer there. I believe we need a strategy to do this in this file, but also for `AtomVariable.cs` and `AtomEvent.cs`. Maybe we can look into [this](https://forum.unity.com/threads/ondestroy-and-ondisable-are-not-called-when-deleting-a-scriptableobject-file.1129220/#post-7259671) and see if we can approach it in a similar fashion. 

Furtermore, just checking for null won't suffice. You can still get this error:
`MissingReferenceException: The object of type 'IntVariable' has been destroyed but you are still trying to access it.` Got the above while testing when I: 
1) Enabled enter play mode in `unity-atoms-examples`. Eg: 
Skärmavbild 2023-08-03 kl  07 00 09
2) Added debug statements  in `HandlePlayModeStateChange` in `AtomVariable.cs` like this: 
```
#if UNITY_EDITOR
        private static void HandlePlayModeStateChange(PlayModeStateChange state)
        {
            Debug.Log($"HandlePlayModeStateChange: {state}. Instances: {_instances.Count}");
            if (state == PlayModeStateChange.ExitingEditMode)  // BEFORE any GO is initialized:
            {
                foreach (var instance in _instances)
                {
                    instance.SetInitialValues();
                    Debug.Log($"HandlePlayModeStateChange instance: {instance.name}");
                }
            }
            else if (state == PlayModeStateChange.EnteredPlayMode)  // within/end of the first frame
            {
                foreach (var instance in _instances)
                {
                    instance.TriggerInitialEvents();
                    Debug.Log($"HandlePlayModeStateChange instance: {instance.name}");
                };
            }
        }
#endif
```

Packages/Core/Runtime/ValueLists/AtomValueList.cs
```diff
@@ -42,6 +44,59 @@ public abstract class AtomValueList<T, E> : BaseAtomValueList, IList<T>
...
+                    if(instance._startCleared) instance.list.Clear();
+                    instance._initial = instance.list.ToList();
+                }
+            }
+            else if (state == PlayModeStateChange.EnteredEditMode) // AFTER Playmode stopped
+            {                
+                foreach (var instance in _instances)
+                {
+                    if (instance == null) // might be for UnityEngine.Objects
```
AdamRamberg, line 90: Same as above

AdamRamberg (changes requested):
When testing and reviewing this PR I realized that we need a way to remove instances in the HashSets for the enter play mode feature. See more info in my comments. What are your thoughts?

Packages/Core/Runtime/Events/AtomEvent.cs
```diff
@@ -62,15 +62,15 @@ private void OnEnable()
 #if UNITY_EDITOR
         private static void HandlePlayModeStateChange(PlayModeStateChange state)
         {
-            if (state == PlayModeStateChange.ExitingEditMode)
+            if (state == PlayModeStateChange.ExitingEditMode) // BEFORE any GO is initialized:
```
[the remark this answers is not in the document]
soraphis: I've had them on purpose separated explicitly. Yes in the case of the AtomValueList in it's current state they are doing the same, but that might not always be the case. (see https://github.com/unity-atoms/unity-atoms/blob/master/Packages/Core/Runtime/Variables/AtomVariable.cs#L193 where they are used to do different things)

they are different events triggered at specific and different points in time/program flow.
that was the reasoning behing making them explicitly two blocks.

But I can change that, if you prefer it compact.

soraphis:
Ok, #421 will be changed to handle dead-references correctly - also for ValueLists.
I made the if statement more compact

AdamRamberg (approved):
LGTM and great that you splitted the work into 2 PRs 👏

Issue #417, opened 2023-07-22
[BUG] NullReferenceException in HandlePlayModeStateChange


burakomer:
**Describe the bug**
When exiting or entering the Play mode, sometimes there are null references left in the `_instances` static list of atom variables inside the AtomVariable generic class. This leads to NullReferenceException when iterating this list on `HandlePlayModeStateChange` method

**To Reproduce**
I don't know when it happens.

**Expected behavior**
No exceptions

**Screenshots**
I don't have any right now.

**Environment**
 - OS: macOS Editor
 - Unity Version 2021.3.21f1
 - Version of Unity Atoms v4.4.5

**Additional context**
Nothing

AdamRamberg:
Fixed in #421

PR #421, opened 2023-07-24
Handling dead references when resetting objects.

soraphis:
Fixes: #417

Objects might be null (as they are unity objects) after an asset is deleted (or an AtomInstancer deleted it's runtime asset) - so these assets would through NRE when resetting with disabled domain reload.

the NRE would also prevent the other objects to reset correctly

### DIFF
--- a/Packages/Core/Runtime/Events/AtomEvent.cs
+++ b/Packages/Core/Runtime/Events/AtomEvent.cs
@@ -76,6 +76,7 @@ namespace UnityAtoms
 
         private void OnDisable()
         {
+            _instances.Remove(this);
             // Clear all delegates when exiting play mode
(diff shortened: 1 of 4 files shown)


soraphis:
This now depends on #420 first, as #420 brings the AtomValueList to the same DDR (disabled domain reload) support as the other objects are.
and then makes all of them handle OnDisable correctly.

AdamRamberg (approved):
I believe this is good enough for our use case ✅

I made the following changes:
- Merged canary
- Added comments to the OnDisable methods

Note though that OnDisable will not be called when deleting the Atom from the editor. Therefore, there might still be null instances, but even though that isn't ideal, it should not cause any problems.

Issue #402, opened 2023-05-25
[BUG] FSM not compatible with disabled domain reload


lgarczyn:
**Describe the bug**
Variables use OnEnable to reset themselves

However, Unity does not call OnEnable on scriptable objects when entering playmode IFF:

* the ScriptableObject is already loaded
* domain reload on play is disabled

**To Reproduce**
1. Disable Domain Reload
2. Create any variable
3. Change the variable value during play mode
4. Value will be preserved on next play mode

**Expected behavior**
Variables should reset to initial value for every play mode sessions

**Additional context**
My current hack
```cs
using UnityAtoms;
using UnityEditor;
using UnityEngine;

namespace Atoms.Editor.EnsureVarReset
{
  // ensure class initializer is called whenever scripts recompile
  [InitializeOnLoad]
  public static class EnsureVarReset
  {
    // register an event handler when the class is initialized
    static EnsureVarReset()
    {
      // Whenever the game enters or leaves playmode
      EditorApplication.playModeStateChanged += LogPlayModeState;
    }

    static void LogPlayModeState(PlayModeStateChange state)
    {
      // Reset each loaded variable
      foreach (AtomBaseVariable variable in ScriptableObject.FindObjectsOfType<AtomBaseVariable>())
      {
        // Reset the var
        variable.Reset();
      }
    }
  }
}
```

soraphis:
which version of atoms are you using?

in 4.4. this #182 should be in place, which works by registering to the playModeStateChanges already

soraphis:
I followed your reproduction steps. For me it works fine. 
Will close the issue. Feel free to re-open it when you have more information about it.

lgarczyn:
I'm on 4.4.5

I only noticed the bug when trying to use FSMs, can you try to reproduce with an FSM instead ?

soraphis:
okay, FSM are infact the problem.

there are multiple layers to this, as it seems:

- OnEnable in AtomVariable is not virtual, so the OnEnable method in the FSM actually overrides the OnEnable of the base class and has no way to invoke it.
- Even after adjusting this, there is a weird behaviour in the FSM that the transition seems to only work every second time (yeah, don't ask me yet why this is the case)

lgarczyn:
> okay, FSM are infact the problem.
> 
> there are multiple layers to this, as it seems:
> 
[...]

PR #422, opened 2023-07-24
make FSM compatible with disabled domain reload 

soraphis:
Fixes: #402

I hope that there is no oversight on my end, but from my testing it seemed to be fine.


AdamRamberg (approved):
Great find 👏 LGTM ✅

PR #432, opened 2023-10-07
Removed stopping play mode in SceneField:Validate()

0xF6:
There is no need to stop the play mode, because there are scenes from Addressables that do not have their own build index.

_and in general it is not convenient, there is enough error~_

### DIFF
--- a/Packages/SceneMgmt/Runtime/Types/SceneField.cs
+++ b/Packages/SceneMgmt/Runtime/Types/SceneField.cs
@@ -87,12 +87,6 @@ namespace UnityAtoms.SceneMgmt
             {
                 /* Sadly its not easy to find which gameobject/component has this SceneField, at least not at this point */
                 Debug.LogError($"A scene [{_sceneName}] you used as reference has no valid build Index", _sceneAsset);
-                // Exit play mode - might be a bit too harsh?
-#if UNITY_2019_1_OR_NEWER
-                EditorApplication.ExitPlaymode();
-#else
-                EditorApplication.isPlaying = false;
-#endif
             }
 #endif
         }


0xF6:
Also, I looked for ways to check the presence of a scene in addressables cases to preserve the intended behavior, but I didn't find anything I needed - so I think that removing the stop of the play mode is the only option

soraphis:
I don't get the purpose of this PR.
from your commit message I guess that it fails for scenes that are from the addressables system, this would be an issue - but the code lines you removed also handle other cases which have nothing to do with addressables.

We can't introduce a "fix" for an optional package, if it introduces issues for the main engine.
So I can't really approve it. There _has_ to be another solution. The easiest would be to create your own kind of SceneVariable that is also allowed to handle Addressable Scenes. 

Sadly I don't know enough about the Addressable scenes to know why those scenes are allowed to have no build index

0xF6:
>Sadly I don't know enough about the Addressable scenes to know why those scenes are allowed to have no build index

Because the build index is an old system, it is needed only for the pack scene into build of the game

>We can't introduce a "fix" for an optional package              

`>optional ` :trollface: 


Well, good luck with the approach of "we will show an error and not let you start the play mode", I'll probably give up this package, there are too many memes like this

soraphis:
- You could run your own fork of the package and do your custom modifications. That would solve your personal needs but keeps the package functioning.
- You could create a custom Atom, instead of using the built-in one. That is one of the core pillars of atoms. Yeah, we try to provide a lot of basic stuff (and some more specialized stuff) from the get go, but we are absolutely aware that different projects are different and it is absolutely encouraged for people to create their own datastructures as needed
- You could provide arguments and alternatives to your case that makes it easier understandable, from observing the code, your change would break a security feature of package

But instead you behave like a small kid.

- You ignore the "contributing.md" which states that a PR must have an issue associated with it (while this is fine for me, it is important because it allows to gather all relevant information for the problem and possible solutions to it, which is one of the reasons why I can't accept your PR as is: there are too many unknowns currently)
- As long as there is no deprecated flag on the buildIndex it is _not_ an old system. 
- by definition the Addressable system IS optional, and no amount of "trollface"-emojis will change that. It might be mandatory for your use case and I'm all in to make atoms work better with Addressables, but not in this destructive manner as this PR suggests.
- If the package is not to your liking, feel free to pick any other SOA package or create your own. You can argue that Unity Atoms is somewhat opinionated and that might not be for everyone. That is fine for me and it seems to be fine for quite a few other people, but I understand if it's not your jam.

Issue #436, opened 2023-10-27
[BUG] QuaternionReferenceDrawer Rendering issue


soraphis:
somewhat related: https://github.com/unity-atoms/unity-atoms/pull/359

# Description

Testing the rendering of different ReferenceDrawer with the following code
```csharp
  public class Issue_353_Test : MonoBehaviour
  {
      public FloatReference floatRef;
      public Vector2Reference float2Ref;
      public Vector3Reference float3Ref;
      public QuaternionReference quatRef;
      
      public IntReference    intRef;
      public StringReference    strRef;
  
      public My3StringStructReference str3Ref;
      public MyStringListStructReference strNRef;
  }
```

Result in unity 2021.3:

| 2021.3 | 2022.2.16 |
| --- | --- |
| ![image](https://github.com/unity-atoms/unity-atoms/assets/3066582/757a13c1-00b2-42cb-aa98-b508dee736c5) |  ![image](https://github.com/unity-atoms/unity-atoms/assets/3066582/bfdd2434-79da-49e2-b403-459c78f2813b)  | 

The issue is that the GetPropertyHeight call here: https://github.com/unity-atoms/unity-atoms/blob/v4.4.8/Packages/Core/Editor/Drawers/AtomBaseReferenceDrawer.cs#L85 does not return the correct value for Quaternions. (see related forum post here: https://forum.unity.com/threads/quaternions-property-height-is-wrong-in-widemode-is-false.1028116/) 

This **is** fixed for the 2022.2.16+ versions, but only in the GetPropertyHeight overload for SerializedPropertyType.
Thats why it is still broken in the second screenshot.

> note that in 2022.2.16+ the rendering for custom structs is broken as the expansion arrows are not placed at the right spot, but this will be addressed in another issue.

AdamRamberg:
Fixed in #440 / #437 and #447

PR #437, opened 2023-10-27
Fixing a Rendering issue of QuaternionReferenceDrawer

soraphis:
due to an unity bug with the result of GetPropertyHeight for Quaternion properties we need to handle Quaternions as special cases

Commit-Type: Bugfix
Resolves: #436 

| 2021.3 | 2022.2.16+ |
| --- | --- |
| image | image |


AdamRamberg:
Fixed in #440

PR #454, opened 2024-08-28
Handling physics disabled

Kyrw:
Created new PR pointing to `canary`

Unity Atoms cant handle physics and physics2d module being disabled on a project. This is a simples (but tedious) enhancement to fix it.

If anyone is going to test it, just uninstall physics module from Package Manager > Build-in and RESTART Unity. This was the only way that I could make sure that the asmdef version define was correctly updated. ( the other way us to update the expression on the asmdef). Same thing for checking if after reinstalling the package everything is working.

The define name was though as PACKAGE_(PARENTNAME)_(PACKAGENAME), skipping the modules.
I.E: com.unity.modules.physics2d > PACKAGE_UNITY_PHYSICS2D. If there is any better patter please let me know.

### DIFF
--- a/Packages/BaseAtoms/Editor/Drawers/Constants/Collider2DConstantDrawer.cs
+++ b/Packages/BaseAtoms/Editor/Drawers/Constants/Collider2DConstantDrawer.cs
@@ -1,4 +1,4 @@
-#if UNITY_2019_1_OR_NEWER
+#if UNITY_2019_1_OR_NEWER && PACKAGE_UNITY_PHYSICS2D
 using UnityEditor;
 using UnityAtoms.Editor;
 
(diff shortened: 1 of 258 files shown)


Kyrw:
Answer from soraphis on the other post

```
https://github.com/unity-atoms/unity-atoms/blob/master/CONTRIBUTING.md
Pull Requests should be made to the canary branch. Is it possible for you to rebase it onto canary?

Also, I guess instead of changing all related files by adding #if PACKAGE_UNITY_PHYSICS2D it would be smarter to create a new asmdef that requires the Physics2D package (adding a define constraint to the asmdef) and move all related files into this asmdef.

should probably do the same for UI related things, for physics3d things, ... and all the other special modules that we have default generations for.

Especially thinking about: most of the changes here are in generated files. so they would break if there would be a need to regenerate those BaseAtoms
```

soraphis:
closing this PR - while it works, i think #455 is a more proper solution, even though it moves way more files around

PR #455, opened 2024-08-28
feat(asmdef): Changed the asmdef structure for more modularity

soraphis:
All physics and physics2D related things now use their own folders and assembly definitions. these are only enabled when the physics modules are loaded and just skipped if not.

this is a direct reference to https://github.com/unity-atoms/unity-atoms/pull/454

using the constraint method of asmdefs:

image


soraphis:
I guess I mistakenly merged this into master oO

PR #459, opened 2024-12-20
modify unity_atom for some outdated code

RequieMa:


soraphis:
thanks for your contribution!
two minor points: 

- should probably use `#if UNTIY_VERSION.....` blocks...
- also the branch should target unity-atoms:canary

soraphis:
I changed the targeted branch and looked into when the property was changed and i think the new one is supported at least since 2021.3 so I guess it should be fine even without the `#if ....` checks

soraphis (approved):

PR #461, opened 2025-05-18
Feature/docusaurus update

soraphis:
updated to docusaurus v3.7
that way i was able to update all the packages, so this should close all current open depandabot PRs:

#394 #382 #381 #380 #378 #377 #354 #338 #336 #335 #334


soraphis:
docs are deployed and running seemingly fine. merging

Issue #365, opened 2022-08-10
[Documentation] Lists/Collections/ValueLists


soraphis:
Add to the docs:

• AtomCollections / AtomLists Tutorials

• AtomValueLists Tutorials

PR #463, opened 2025-05-26
feat: ValueList Docs and Inspector Improvements

soraphis:
feat: Added custom drawer for GameObjectValueList, that way it does not show "Type Missmatch" anymore when referencing runtime Gameobjects.
feat: started adding documentation and a few video clips for ValueLists

I created a custom drawer to fix the "type missmatch" labels in GameObjectValue lists when referencing scene objects in an asset:

<details name="accordion">
 <summary> old visuals </summary>

grafik

</details>

<details name="accordion">
 <summary> new visuals </summary>

grafik

</details>

it supports click-to-ping. it supports drag-and-drop* (but only non-scene objects) and shows the object picker 
- albeit most of it is just "fake" in a way. Unity still does not like referencing runtime objects, so it will make them "null" on drag and drop and stuff. But I'd say it's still better than before

partially addresses: #365

### DIFF
--- a/Packages/BaseAtoms/Editor/Editors/ValueLists/GameObjectValueListEditor.cs
+++ b/Packages/BaseAtoms/Editor/Editors/ValueLists/GameObjectValueListEditor.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using UnityAtoms.BaseAtoms;
+﻿using UnityAtoms.BaseAtoms;
 using UnityEditor;
 using UnityEditorInternal;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Packages/BaseAtoms/Editor/Editors/ValueLists/GameObjectValueListEditor.cs
```diff
@@ -0,0 +1,164 @@
...
+            list.drawHeaderCallback = (Rect rect) =>
+            {
+                EditorGUI.LabelField(rect, serializedObject.FindProperty("list").displayName);
+            };
+            list.drawElementCallback = (Rect rect, int index, bool isActive, bool isFocused) =>
+            {
+                var element = list.serializedProperty.GetArrayElementAtIndex(index);
+                if (element.objectReferenceValue == null)
+                {
+                    EditorGUI.PropertyField(rect, element, GUIContent.none);
```
Kyrw, line 36: Maybe add a return and remove the else. using a if escape. Just for better readibilty

docs/assets/valuelist/valuelist_02_create_and_sync.mp4
Kyrw, line 1: Should we be adding videos to the repo? Eventhough it's on the docs folder?

Kyrw (approved):
Just commented on some suggetions, but everything seems nice

PR #494, opened 2026-03-04
Update SetVariableValue.cs

builder-main:
Allow to force event on all set variable value. Should be backward compatible


soraphis (approved):
- this is a logical follow up on https://github.com/unity-atoms/unity-atoms/issues/131
- does not introduce breaking change
- Setting of the Value property also called SetValue(...., false) internally

PR #499, opened 2026-05-23
Merge with original package

Ferhattepe:


soraphis:
briefly watched over the code. All in all i have nothing against these changes, but:

- PRs should target canary, not master.
- Some breaking changes in here (renaming of properties), we can add properties, but cannot easily remove them.
- A lot of stuff grouped together (e.g. the editor window changes) that should each receiver their own PR

